PR #54, opened 2018-12-20
fix to data mapping in OpenMP target version of TRAP_INT kernel

mcordery:
Fixes a data mapping problem in TRAP_INT-OMPTarget where a data environment is created via

#pragma target  data map

rather than an async mapping via

#pragma target enter data map

This causes a failure at the subsequent exit data map. Fixed the mapping to use the  target
enter data map and kernel runs fine.


dmcdougall:
This patch fixes the same problem for me as well.

dmcdougall:
Is there any update on this PR?

dmcdougall:
Another gentle ping :)

davidbeckingsale:
Thanks for the reminder @dmcdougall

PR #64, opened 2019-04-17
Add Caliper Reporting

DavidPoliakoff:
This adds Caliper in a very light fashion to the code, instrumenting metadata features of the run and using the reporting functionality of RAJAPerf to tell Caliper about the performance.

As a prerequisite for this, rajaperf_config.hpp wasn't playing nicely with PGI, I changed how that worked, initializing its values in a source file rather than the header.

First, if you don't explicitly enable Caliper, very few changes happen, you have another submodule initialization and a few empty function calls.

If you do enable Caliper, the reporting part of RAJAPerf "declares" its results, which Caliper then consumes. I'll attach some screenshots to show what you can do with this in the comments


DavidPoliakoff:
Results: 

First, I built the perfsuite with a bunch of compilers (PGI, GNU, Clang, I'm working on Intel and IBM). I spit out results as a JSON file, which I imported into a Jupyter notebook. After doing some Pandas stuff to the results, I wound up with the following features: Kernels, Timings per variant, Compilers, and Compiler options. Further, I added "% of time PGI takes as compared to Clang" for both RAJA and Base variants (100% means PGI is as fast as Clang, 1000% means it's 10x slower, 10% means it's 10x faster).

This leads to a bit of a noisy file

[Noisy Image](https://imgur.com/a/CoSvSX2)

Which we can then filter for interesting results. It turns out for a lot of Polybench, the Base variants absolutely scream with PGI, but the RAJA variant is at least 3x slower

[Filtered down to interesting results](https://imgur.com/a/8hZXwHQ)

PR #107, opened 2021-02-09
Always allow Base_Seq variants to run

### LINKED ISSUE #106: Fall-through case statements in KernelBase::runKernel cause Base_Seq to not run if RAJA_Seq can't run
The logic in these lines of KernelBase::runKernel uses fall-through case statements to run all the Sequential variants off a function. When `RUN_RAJA_SEQ` is not defined none of the sequential cases can be run.

https://github.com/LLNL/RAJAPerf/blob/5122ad1be05b8fa67fbebaebb581ae4a8b304f38/src/common/KernelBase.cpp#L97-L105

MrBurmark:
Always allow Base_Seq variants to run. Unlike the parallel variants which are enabled or disabled based on Raja support for the backed base sequential variants are always enabled and will now always be allowed to run.

Fixes #106.

### DIFF
--- a/src/common/KernelBase.cpp
+++ b/src/common/KernelBase.cpp
@@ -95,6 +95,11 @@ void KernelBase::runKernel(VariantID vid)
   switch ( vid ) {
 
     case Base_Seq :
+    {
+      runSeqVariant(vid);
+      break;
+    }
+
     case Lambda_Seq :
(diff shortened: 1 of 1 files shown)


rhornung67 (approved):
Wow. I don't know how I screwed that up. Thanks for the fix.

PR #126, opened 2021-06-05
Feature/burmark1/per loop timings

MrBurmark:
Add an output file for timing on a per loop/kernel basis. This mainly helps us understand how close to being launch overhead bound we are with gpu backends.


MrBurmark:
@rhornung67 Timing per kernel can be calculated from the timings file along with the number of repetitions and loops per repetition from the run resport. Should a timings per loop CSV file be created or should we leave that up to the user to calculate?

MrBurmark:
included in #147

PR #139, opened 2021-06-18
Default Kernel Sizes to 1M

MrBurmark:
Change the default kernel size to about 1M.
Most kernels work on 1D arrays of the problem size so their default size was changed to 1M and their default repetitions scaled appropriately.
Some kernels work on 2D or 3D data. For the most part multidimensional kernels are changed to scale linearly in total size by scaling their dimensions evenly as a square or cube. This is a departure as some kernels previously scaled by changing a single dimension. Some other kernels scaled each dimension by the size factor creating quadratic or cubic total size scaling.
The LTIMES default kernel size and scaling is not changed.
The polybench kernels will be addressed in a separate PR.

### DIFF
--- a/src/algorithm/SORT.cpp
+++ b/src/algorithm/SORT.cpp
@@ -22,7 +22,7 @@ SORT::SORT(const RunParams& params)
   : KernelBase(rajaperf::Algorithm_SORT, params)
 {
    setDefaultSize(1000000);
-   setDefaultReps(50);
+   setDefaultReps(20);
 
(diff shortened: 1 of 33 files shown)


rhornung67 (approved):
We should think about how to do something similar for nested loop kernels.

MrBurmark:
For the 2d and 3d loops here I knew how to make a problem size of 1M even if the number of iterations wasn't 1M. For some of the nested loops we may need to ask the originator how to do something similar.

PR #146, opened 2021-06-25
FLOPs per rep

MrBurmark:
Add FLOPs per rep, with FLOP generously meaning anything from an add to multiply, to a divide, sqrt, or exponential.

### DIFF
--- a/src/algorithm/SORT.cpp
+++ b/src/algorithm/SORT.cpp
@@ -24,6 +24,8 @@ SORT::SORT(const RunParams& params)
   setDefaultSize(1000000);
   setDefaultReps(20);
 
+  setFLOPsPerRep(0);
+
   setUsesFeature(Sort);
 
   setVariantDefined( Base_Seq );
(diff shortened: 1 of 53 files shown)


artv3:
Out of curiosity have you compared with what nvprof reports for flops? To see if the numbers are close?

rhornung67 (commented):
I believe this looks OK. It may be a good idea for a small group of us to walk through changes like this to see if we agree on the flop counts.  Also, we should sequence PRs like this to avoid massive merge conflicts. I'm re-working the problem size and its per rep stuff.

rhornung67 (approved):
I'll approve now. We can get more eyeballs on it later. I'll wait until this is merged before I finish my rework changes.

MrBurmark:
You're right about merge conflicts. I probably should have waited, but I had some free energy last night so I just jumped into this.

MrBurmark:
@artv3 That's a good idea, let me do a quick sanity check with nsight.

artv3 (approved):

PR #148, opened 2021-06-29
Update ENERGY kernels per rep

MrBurmark:
Fix energy kernels per rep to be 6 instead of 1.


rhornung67 (approved):
Thanks. I thought I checked everything for correctness. Guess not.

PR #149, opened 2021-06-29
Update GEN_LIN_RECUR kernels per rep

MrBurmark:
Fix lcals GEN_LIN_RECUR kernels per rep to be 2 instead of 1


rhornung67 (approved):
Thanks.

PR #166, opened 2021-07-27
Fix kernel arguments in mat mat multiply with shared memory

artv3:
The kernel arguments were out of order leading to incorrect output.


rhornung67 (approved):
Thanks @artv3

PR #169, opened 2021-08-03
Use one buffer in REDUCE3_INT Base variants

MrBurmark:
Optimize REDUCE3_INT cuda and hip variants so they only use a single buffer. This reduces the number of copies improving performance.


artv3 (approved):
Nice! How much faster did it get?

MrBurmark:
About 1.25x to 3x depending on problem size.

rhornung67 (approved):

PR #186, opened 2021-09-07
Use threadIdx.x rather than legacy HIP thread Id calls in MASS3DPA-HIP. 

artv3:


rhornung67 (approved):
Does this change make any difference in performance?

artv3:
> Does this change make any difference in performance?

Nope, numbers look about the same.

PR #188, opened 2021-09-16
Feature/kokkos

ajpowelsnl:
# Summary

@davidbeckingsale and @rhornung67  have been our main points of contact and collaboration in this work

The proposed code implements RAJAPerf Suite (RPS) - based performance testing for Kokkos.

Main new code and changes include: 

- Bringing Kokkos in as a RPS (tpl) submodule
- Kokkos implementations (with Kokkos_Lambda as a kernel variant) of the basic, stream, lcals , apps and algorithm kernel groups (polybench will be completed in FY22)
- Changes to the build system / cmake
- Significant changes to the infrastructure for registering new kernel groups and new kernels:

common/Executor.hpp:  kernelID registerKernel(std::string, KernelBase*);
common/Executor.cpp:        exec->registerKernel(groupName, kernel);
RAJAPerfSuiteDriver.cpp:  //executor.registerKernel

# Configure Kokkos build and runtime in RPS:
cmake \
-DENABLE_KOKKOS=ON \
-DENABLE_OPENMP=ON \
-DCMAKE_BUILD_TYPE=Debug .. \

Please do not hesitate to contact @ajpowelsnl or @DavidPoliakoff for any questions and / or further discussion.


CMakeLists.txt
```diff
@@ -25,9 +25,8 @@ set(ENABLE_TESTS Off CACHE BOOL "Enable BLT and RAJA tests")
 set(ENABLE_EXAMPLES Off CACHE BOOL "Enable RAJA examples")
 set(RAJA_ENABLE_EXERCISES Off CACHE BOOL "Enable RAJA exercises")
 
-set(CMAKE_CXX_STANDARD 11)
-set(BLT_CXX_STANDARD 11)
-
+set(CMAKE_CXX_STANDARD 14)
```
davidbeckingsale, line 28: Does Kokkos require C++14? If so, let's make this conditional depending on `ENABLE_KOKKOS`

CMakeLists.txt
```diff
@@ -80,16 +78,19 @@ if (ENABLE_OPENMP)
 endif()
 if (ENABLE_CUDA)
   list(APPEND RAJA_PERFSUITE_DEPENDS cuda)
-endif() 
-if (ENABLE_HIP)
+endif()
+# HIP is used with AMD / VEGA GPU
+# Neatly separate RAJAPerf Suite and Kokkos handling of HIP
```
davidbeckingsale, line 83: Maybe expand with a bit more detail here. IIRC it was because Kokkos requires hipcc as the CMAKE_CXX_COMPILER, whereas RAJA uses blt/CMake FindHIP

CMakeLists.txt
```diff
@@ -80,16 +78,19 @@ if (ENABLE_OPENMP)
...
+#if (ENABLE_HIP)
   list(APPEND RAJA_PERFSUITE_DEPENDS hip)
 endif()
 
 set(RAJAPERF_BUILD_SYSTYPE $ENV{SYS_TYPE})
 set(RAJAPERF_BUILD_HOST $ENV{HOSTNAME})
 
 if (ENABLE_CUDA)
-  set(CMAKE_CUDA_STANDARD 11)
+  set(CMAKE_CUDA_STANDARD 14)
```
davidbeckingsale, line 93: Same comment as the C++ standard stuff - I think this should be conditional.

CMakeLists.txt
```diff
@@ -113,7 +114,75 @@ configure_file(${CMAKE_SOURCE_DIR}/src/rajaperf_config.hpp.in
 # remove project-specific CMake variables that are no longer needed)
 set (CUDA_NVCC_FLAGS ${RAJA_NVCC_FLAGS})
 
+# HACKS TO FIX COMPILATION ISSUES
+include_directories(SYSTEM ${CMAKE_CURRENT_SOURCE_DIR}/tpl/RAJA/include/)
```
davidbeckingsale, line 118: Is this always required, or only when Kokkos is enabled?

src/basic/INIT3-OMPTarget.cpp
```diff
@@ -80,10 +80,10 @@ void INIT3::runOpenMPTargetVariant(VariantID vid)
     startTimer();
     for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
 
-      RAJA::forall<RAJA::omp_target_parallel_for_exec<threads_per_team>>(
-        RAJA::RangeSegment(ibegin, iend), [=](Index_type i) {
-        INIT3_BODY;
-      });
+      //RAJA::forall<RAJA::omp_target_parallel_for_exec<threads_per_team>>(
```
davidbeckingsale, line 83: ??

src/basic/INIT_VIEW1D-OMPTarget.cpp
```diff
@@ -72,10 +72,10 @@ void INIT_VIEW1D::runOpenMPTargetVariant(VariantID vid)
      startTimer();
      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
 
-       RAJA::forall<RAJA::omp_target_parallel_for_exec<threads_per_team>>(
-         RAJA::RangeSegment(ibegin, iend), [=](Index_type i) {
-         INIT_VIEW1D_BODY_RAJA;
-       });
+       //RAJA::forall<RAJA::omp_target_parallel_for_exec<threads_per_team>>(
```
davidbeckingsale, line 75: ??

davidbeckingsale (changes requested):
Just looked at a few files to start with. Looks like there's a few TODO comments around, and commented code that might need to be removed.

artv3:
@ajpowelsnl did you branch from an older version of the persuite. I was interested in looking at the kokkos version of the `MAT_MAT_SHARED` kernel, but don't see it.

CMakeLists.txt
```diff
@@ -80,18 +79,23 @@ if (ENABLE_OPENMP)
...
 if (ENABLE_CUDA)
   list(APPEND RAJA_PERFSUITE_DEPENDS cuda)
-endif() 
-if (ENABLE_HIP)
+endif()
+# Kokkos requires hipcc as the CMAKE_CXX_COMPILER for HIP AMD/VEGA GPU
+# platforms, whereas RAJAPerf Suite uses blt/CMake FindHIP to set HIP compiler 
+# Separate RAJAPerf Suite and Kokkos handling of HIP compilers
+if ((ENABLE_HIP) AND (NOT ENABLE_KOKKOS))
+#if (ENABLE_HIP)
```
davidbeckingsale, line 87: This line can be removed.

CMakeLists.txt
```diff
@@ -80,18 +79,23 @@ if (ENABLE_OPENMP)
...
   list(APPEND RAJA_PERFSUITE_DEPENDS hip)
 endif()
 
 set(RAJAPERF_BUILD_SYSTYPE $ENV{SYS_TYPE})
 set(RAJAPERF_BUILD_HOST $ENV{HOSTNAME})
 
 if (ENABLE_CUDA)
-  set(CMAKE_CUDA_STANDARD 11)
+  if (ENABLE_CUDA AND ENABLE_KOKKOS)
```
davidbeckingsale, line 95: The ENABLE_CUDA here is redundant (since we are already inside an `if (ENABLE_CUDA)` block.

CMakeLists.txt
```diff
@@ -106,14 +110,85 @@ else()
...
+	  set(Kokkos_ARCH_VOLTA70 ON CACHE BOOL "Docstring") #TODO: better
+    enable_language(CUDA)
+  endif()
+  if(ENABLE_OPENMP)
+    #set(Kokkos_ENABLE_OPENMP CACHE BOOL ON)
+    set(Kokkos_ENABLE_OPENMP ON CACHE BOOL "Docstring")
+  endif()
+
+  add_subdirectory(tpl/kokkos)
+  if(ENABLE_CUDA)
```
davidbeckingsale, line 170: This block is empty.

CMakeLists.txt
```diff
@@ -106,14 +110,85 @@ else()
...
+	  #        get_source_file_property(local ${kokkos_core_file} LANGUAGE) 
+	  #        set(remote "")
+	  #        #get_source_file_property(remote ${kokkos_core_file} DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/tpl/kokkos/core/src LANGUAGE) 
+	  #        message(STATUS "Kokkos file: ${kokkos_core_file} ${local} ${remote}")
+	  #        
+	  #endforeach()
+    #set_target_properties(kokkoscore kokkoscontainers PROPERTIES LANGUAGE CUDA)
+  endif()
+  get_property(KOKKOS_INCLUDE_DIRS DIRECTORY tpl/kokkos PROPERTY INCLUDE_DIRECTORIES)
+  include_directories(${KOKKOS_INCLUDE_DIRS})
```
davidbeckingsale, line 185: Is this required? I think that having `kokkos` in the RAJA_PERFSUITE_DEPENDS should be enough.

src/CMakeLists.txt
```diff
@@ -136,45 +188,57 @@ blt_add_executable(
...
+  lcals-kokkos/FIRST_DIFF-Kokkos.cpp
+  lcals-kokkos/FIRST_MIN-Kokkos.cpp
+  lcals-kokkos/FIRST_SUM-Kokkos.cpp
+  lcals-kokkos/GEN_LIN_RECUR-Kokkos.cpp
+  lcals-kokkos/HYDRO_1D-Kokkos.cpp
+  lcals-kokkos/HYDRO_2D-Kokkos.cpp
+  lcals-kokkos/INT_PREDICT-Kokkos.cpp
+  lcals-kokkos/PLANCKIAN-Kokkos.cpp
+  lcals-kokkos/TRIDIAG_ELIM-Kokkos.cpp
+  #polybench/POLYBENCH_2MM.cpp
```
davidbeckingsale, line 203: Why are these all commented?

src/RAJAPerfSuiteDriver.cpp
```diff
@@ -7,14 +7,38 @@
...
+            Kokkos::initialize(argc, argv);
+#endif // RUN_KOKKOS
+
   rajaperf::Executor executor(argc, argv);
+  rajaperf::make_perfsuite_executor(&executor, argc, argv);
+  //executor.registerKernel
+  //rajaperf::RunParams params(argc, argv);
+  //executor.registerGroup("Sparse");
+
+  //executor.registerKernel("Sparse", rajaperf::make_kernel_base(
```
davidbeckingsale, line 29: Commented code?

src/apps/WIP-COUPLE.cpp
```diff
@@ -20,184 +20,175 @@ namespace rajaperf
 namespace apps
 {
 
+//
+//COUPLE::COUPLE(const RunParams& params)
```
davidbeckingsale, line 24: Lots more commented code?

davidbeckingsale (changes requested):

src/CMakeLists.txt
```diff
@@ -8,27 +8,53 @@
 
 include_directories(.)
 
+# Added as part of the merge with build only infrastructure:
 add_subdirectory(common)
+
+if(NOT INFRASTRUCTURE_ONLY)
 add_subdirectory(apps)
+add_subdirectory(apps-kokkos)
```
davidbeckingsale, line 16: I think these should be guarded with ENABLE_KOKKOS

src/CMakeLists.txt
```diff
@@ -8,27 +8,53 @@
 
 include_directories(.)
 
+# Added as part of the merge with build only infrastructure:
 add_subdirectory(common)
+
+if(NOT INFRASTRUCTURE_ONLY)
```
davidbeckingsale, line 14: Where does the INFRASTRUCTRUE_ONLY option come from?

src/CMakeLists.txt
```diff
@@ -190,24 +254,33 @@ blt_add_executable(
...
   common/RunParams.cpp
-  algorithm/SORT.cpp
-  algorithm/SORT-Seq.cpp
-  algorithm/SORTPAIRS.cpp
-  algorithm/SORTPAIRS-Seq.cpp
   DEPENDS_ON ${RAJA_PERFSUITE_DEPENDS}
 )
 
 else()
+	if(NOT INFRASTRUCTURE_ONLY)
```
davidbeckingsale, line 279: This looks unnecessary.

davidbeckingsale (changes requested):

src/CMakeLists.txt
ajpowelsnl: > Where does the INFRASTRUCTRUE_ONLY option come from?

Hi David -- Taking a step back to explain the "big picture", remember that we are using RAJAPerf Suite (RPS) in two distinct manners:  1) Provide Kokkos translations of RPS kernels that can be run along side native RPS kernels, and 2) Using RPS as the **_infrastructure driver_** for Kokkos Kernels and Kokkos performance tests that we have written.  To be clear, this second case has nothing to do with RPS kernels & their Kokkos translations.   This second case is tied to the "INFRASTRUCTURE_ONLY" option.  Does that make sense?

src/CMakeLists.txt
ajpowelsnl: > I think these should be guarded with ENABLE_KOKKOS

Hi David, if I am understanding your suggestion, guarding all things Kokkos in this file will require significant reorganization of this file.  Is that what you would like for me to do, i.e., cleanly separate RAJAPerf Suite content from Kokkos content?

src/CMakeLists.txt
ajpowelsnl: Hi David -- Please see my comment above on this point (i.e., there are two distinct scenarios in which we're using RAJAPerf Suite).  Would it be possible for us to meet, and work through at least some / most of your suggested changes?  I just want to make sure I'm doing what you want me to do for the merge.

src/CMakeLists.txt
davidbeckingsale: The _why_ makes sense, I meant that INFRASTRUCTURE_ONLY should be explicitly defined as a CMake `option` (https://cmake.org/cmake/help/latest/command/option.html)  with a default value of False, that you can override when you are using the perfsuite as the infrastructure driver.

CMakeLists.txt
ajpowelsnl: DONE!  Apologies, I'm circling back around to your comments, and I'm just now seeing this - apologies for the oversight!

CMakeLists.txt
ajpowelsnl: I deleted the if (ENABLE_CUDA) and endif() bloc.

src/CMakeLists.txt
ajpowelsnl: The Polybench kernels are commented because we do not yet have Kokkos translations for them.  I believe the situation calls for additional logic to deal with the relative immaturity of the Kokkos design.  Does this make sense?  It's a choice David P and I made in the course of prototyping.

src/RAJAPerfSuiteDriver.cpp
ajpowelsnl: Deleted all debugging code.

src/common/Executor.cpp
```diff
@@ -158,15 +165,18 @@ void Executor::setupSuite()
     run_params.setInvalidExcludeFeatureInput(invalid);
 
     //
```
ajpowelsnl, line 167: TODO:  Rich, David B., there is functionality we have not yet implemented in Kokkos, but that is important for RAJAPerf Suite.  Starting at line 167, and ending at ~ line 342 in Executor.cpp, we have commented this functionality.  Rich, David B. suggested that I work with you on how to handle integrating this (currently) commented functionality with Kokkos design (i.e., dynamic addition of new kernels and kernel groups).

ajpowelsnl, line 167: **TODO**: @rhornung67 , @davidbeckingsale , there is functionality we have not yet implemented in Kokkos, but that is important for RAJAPerf Suite. Starting at line 167, and ending at ~ line 342 in Executor.cpp, we have commented this functionality. Rich, David B. suggested that I work with you on how to handle integrating this (currently) commented functionality with Kokkos design (i.e., dynamic addition of new kernels and kernel groups).

src/common/RunParams.cpp
```diff
@@ -651,24 +642,16 @@ void RunParams::printFeatureKernels(std::ostream& str) const
...
-         delete kern;
-      }
-    }  // loop over kernels
+
     str << std::endl;
   }  // loop over features
   str.flush();
 }
-
+// TODO for Kokkos Team:  Commenting function body, because this infrastructure
```
ajpowelsnl, line 650: TODO:  @rhornung67 , @davidbeckingsale - commented lines here, 654-674, also represent important functionality for RAJAPerf Suite that has not yet been implemented for Kokkos.  Would you guys please help me work through integrating these bits?

crtrott:
Hey guys, I didn't yet read all the comments but I think we should try and split the infrastructure changes from the Kokkos kernel variant changes. 

So essentially something like this:
PR1: Infrastructure changes (how to add Kernel groups etc. + let perf suite serve as a perf testing infrastructure in other projects [i.e. Kokkos Core/Kokkos Kernels])
PR2: Add optional Kokkos dependency with all the kernels variants.

@davidbeckingsale what do you think about this?

Note that PR2 are not really dependent on PR1 so we could also do those two first and then tackle the infrastructure change. We could also split PR2 into doing kernel groups individually I guess?

davidbeckingsale:
> @davidbeckingsale what do you think about this?

I agree that splitting the changes up into two distinct pieces would be best, and I suggested that approach before. I think the two main sticking points on this PR right now are 1) ensuring that the infrastructure changes maintain all the current capability, and 2) deciding how to handle cases where a variant is "missing".

ajpowelsnl:
> > @davidbeckingsale what do you think about this?
> 
> I agree that splitting the changes up into two distinct pieces would be best, and I suggested that approach before. I think the two main sticking points on this PR right now are 1) ensuring that the infrastructure changes maintain all the current capability, and 2) deciding how to handle cases where a variant is "missing".

Hi David & Christian -- Many thanks for responding.  As for sticking point 2, in the most recent push, I did stub in Kokkos variants for the polybench kernels, somewhat alleviating that issue.  But we are still left with sticking point 1.  I will work with Christian on this, and see if we can come up with a satisfactory solution that preserves current infrastructure  capability.

PR #194, opened 2021-10-04
Fix HEAT_3D index mapping

MrBurmark:
# Fix HEAT_3D index mapping

Reorder the hip and cuda RAJA variant index ordering to
get coalesced loads and stores

- This PR is a bugfix
  - Fixes HEAT_3D performance issue


artv3 (approved):

rhornung67 (approved):
D'oh! Good catch. Thanks @MrBurmark

PR #197, opened 2021-10-20
GPU Block Size Support

MrBurmark:
# Support Multiple GPU Block Sizes
This is a first pass at adding gpu block size support. It only supports multiple block sizes for kernels using forall and kernel. Other kernels work but report their block sizes report as 0.
Which block sizes are used is a configure time option that can be changed by setting `RAJA_PERFSUITE_GPU_BLOCKSIZES` to a comma separated list of block sizes like `128,256,512,1024`. If `RAJA_PERFSUITE_GPU_BLOCKSIZES` is not set each kernel uses its default block size, usually 256.
The `--gpu_block_size <number>` argument can be used to run a specific block size for any kernel supporting that block size. If  `--gpu_block_size` is not specified the default block sizes if the first argument to `RAJA_PERFSUITE_GPU_BLOCKSIZES` or the default block size if empty.

NOTE: Please review by looking at each commit one by one.
 - The first commit adds a cmake configure option to RAJAPerf for a list of gpu block sizes via the cmake variable RAJA_PERFSUITE_GPU_BLOCKSIZES. Each kernel supporting multiple block sizes is then built with each of the block sizes in that list. 
 - The next commit adds support for the `--gpu_block_size <number>` argument. This allows you to specify what gpu block size to run. Note that only one block size can be run at a time, if none is specified each kernel uses its own default block size.
 - The next commit adds some utility classes and functions that enable kernels to implement multiple block sizes easily with little extra code. This currently uses mostly custom templates, but it may be possible to use std or camp library function templates to simplify this code.
 - The next commit adds HALO_EXCHANGE_FUSED to the warmup kernels to startup the RAJA pinned pool.
 - The next commit fixes a bug in the GPU implementation of GEN_LIN_RECURS
 - The next commit implements multiple block size support in the DOT and HEAT_3D kernels. This is an example of how kernels using forall (DOT) and kernel (HEAT_3D) can support multiple block sizes.
 - The next commit templates the gpu kernels in DOT and HEAT_3D so they can take advantage of compile time knowledge of the block size.
 - The following pairs of commits add support for multiple block sizes and template the gpu kernels for each directory.

- This PR is a feature
- It does the following:
  - Adds GPU block size support at the request of multiple people


MrBurmark:
In looking through this I realized that in our multidimensional kernels we never changed the hip variants to use 64 instead of 32 threads. So I guess we can do that as part of this or separately. What do people prefer?

MrBurmark:
I saw better performance in multi dimensional kernels when I made the block size in the x direction be the wavefront size. So I would use 64 instead of 32 in teh hip kernels, but I would probably keep 256 threads total.

rhornung67:
@MrBurmark I see what you mean now by changing the hip variants to use 64 threads vs. 32 after looking at the code.

trws:
Yeah, the plan for the 14 transition is to ditch anything below 8 anyway.  Maybe we can fold in the update to the alternate dockerfile and bump everything up.
________________________________

rhornung67:
@MrBurmark is this PR superseded by this one? https://github.com/LLNL/RAJAPerf/pull/212

MrBurmark:
Yes, closing this in favor of #212 .

PR #202, opened 2021-11-02
Optionally output npasses min,max,ave timing

MrBurmark:
# Add option to output min and max as well as average

Have the option to output the min, max, and/or average of
timing data using --npasses-combiner argument. Timing and speedup
csv files append the combiner method to their filename and print
the combiner method in the file.

- This PR is a feature
- It does the following:

### DIFF
--- a/src/common/Executor.cpp
+++ b/src/common/Executor.cpp
@@ -883,8 +883,7 @@ void Executor::writeCSVReport(const string& filename, CSVRepMode mode,
     //
     // Print title line.
     //
-    file << getReportTitle(mode);
-    file << RunParams::CombinerOptToStr(combiner);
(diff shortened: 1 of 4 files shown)


artv3:
Just a though, but for MPI I know caliper can provide this information. One option could be to rely on caliper for more detailed timing info.

MrBurmark:
We're already collecting this info, we just weren't outputting it. I agree caliper really is the place for more detailed stats.

rhornung67 (approved):
@MrBurmark I'm OK with this. Do you want to merge it?

MrBurmark:
Sure, I'll merge it.
If we decide to do something more complicated for MPI later then we can figure out how to combine multi-pass with multi-rank data at that time.

PR #205, opened 2021-12-15
Update threading strategy for Diffusion 3D PA kernel

artv3:
Increases number of threads used in the kernel to increase performance on GPUs, credit to @tomstitt .

### DIFF
--- a/src/apps/DIFFUSION3DPA-Cuda.cpp
+++ b/src/apps/DIFFUSION3DPA-Cuda.cpp
@@ -22,126 +22,100 @@
 namespace rajaperf {
 namespace apps {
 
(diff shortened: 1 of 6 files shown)


rhornung67 (approved):
Interesting. Is there interest in ways to make changes such as this without touching so many lines of source code? For me, that seems to be a strong distinction between RAJA kernel and RAJA teams.

artv3:
> Interesting. Is there interest in ways to make changes such as this without touching so many lines of source code? For me, that seems to be a strong distinction between RAJA kernel and RAJA teams.

Yes absolutely, I think it could be done if the code was written with a bit more generality.  The big change here was going from 2->3D thread blocks, this resulted in the loop bodies and amount of register space needed per thread to change as well going from stack arrays to scalars.

PR #245, opened 2022-05-12
Add script to make graphs from problem size sweeps

MrBurmark:
# Simplify creating graphs from problem size sweeps

Add a script that can create graphs from the output of our problem size sweep script.

- This PR is a feature
- It does the following:
  - Adds graphing script at the request of people who like graphs


artv3:
YES! If its ready do you want to mark it ready for review?

rhornung67 (changes requested):
Looks good. Please add a paragraph in the README.md file about running sweeps and the fact that these scripts exist to make plots.

MrBurmark:
There are still some rough edges in this script and I want to add some more comments.

PR #246, opened 2022-05-20
hip mfma tests

CRobeck:
This PR adds basic functionality test of leveraging the matrix cores on AMD gfx908 and gfx90a hardware for dense matrix products.


src/basic/MAT_FUSED_MUL_ADD-Hip.cpp
```diff
@@ -0,0 +1,308 @@
...
+	std::string hipArch = getArch();
+#if defined(RP_USE_DOUBLE)
+	if (hipArch=="gfx90a") 
+		return true;
+#endif
+#if defined(RP_USE_FLOAT)
+	if (hipArch=="gfx90a" || hipArch=="gfx908")
+		return true;
+#endif
+return false;
```
MrBurmark, line 239: spacing

src/basic/MAT_FUSED_MUL_ADD-Hip.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+    startTimer();
+    RAJA::RangeSegment ii_range(0, (N/(Ne*Ne)));
+    RAJA::RangeSegment row_range(0, Ne);
+    RAJA::RangeSegment col_range(0, Ne);
+
+    using EXEC_POL =
+      RAJA::KernelPolicy<
+        RAJA::statement::HipKernel<
+        RAJA::statement::For<2, RAJA::loop_exec,
```
CRobeck, line 193: @MrBurmark Is there an easy RAJA way to make this loop_exec just go over these extents in parallel?

rhornung67: @CRobeck You should be able to change it to hip_block_z_loop.  

Also, you should probably change the 'loop' policies in line 194 and 195 to 'direct' policies for performance; i.e., 'hip_block_y_loop' --> 'hip_block_y_direct'. Typically, we map a loop over tiles to block-direct policies and a loop withing a tile to a thread-direct policy.

src/basic/MAT_FUSED_MUL_ADD-Hip.cpp
```diff
@@ -0,0 +1,308 @@
...
+#ifdef __gfx908__
+#if defined(RP_USE_FLOAT)
+    result = __builtin_amdgcn_mfma_f32_16x16x4f32(a, b, result, 0, 0, 0);
+#endif  
+#endif
+    a_idx += 4; // move two columns to the right
+    b_idx += 4*Ne; // move two rows down
+  }
+
+  #pragma unroll 4
```
artv3, line 66: We have RAJA macros for unrolling with a fixed amount: 

https://github.com/LLNL/RAJA/blob/develop/include/RAJA/config.hpp.in#L572

src/basic/MAT_FUSED_MUL_ADD-Hip.cpp
rhornung67: Here's an example from another kernel in the perf suite: https://github.com/LLNL/RAJAPerf/blob/develop/src/polybench/POLYBENCH_HEAT_3D-Hip.cpp#L179

src/basic/MAT_FUSED_MUL_ADD-Hip.cpp
```diff
@@ -0,0 +1,301 @@
...
+}
+void MAT_FUSED_MUL_ADD::runHipVariantBuiltin(VariantID vid)
+{
+  const Index_type run_reps = getRunReps();
+  const Index_type iend = getActualProblemSize();  
+  const Index_type N = m_N;
+  constexpr Index_type Ne = m_Ne;
+  constexpr Index_type NeNe = m_Ne * m_Ne;
+
+  dim3 gridDim (1, 1, 1);
```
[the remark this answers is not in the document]
CRobeck: We're using 4 groups of 16 threads for each outer product, so we only need single block per grid.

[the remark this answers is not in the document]
MrBurmark: I'm worried that once we saturate flops for a CU we'll be leaving flops on the table.

src/common/HipDataUtils.hpp
```diff
@@ -178,6 +177,14 @@ void deallocHipPinnedData(T& pptr)
   pptr = nullptr;
 }
 
+static std::string getHipArch()
+{  
+  hipDeviceProp_t devProp;
+  hipGetDeviceProperties(&devProp, 0);
+  std::string gcnArchName(devProp.gcnArchName);
+  std::string hipArch = gcnArchName.substr(0, 6);
```
MrBurmark, line 185: is 0, 6 the right thing for all architectures, aren't there 7 digit gpu names like gfx10##?

src/common/HipDataUtils.hpp
```diff
@@ -22,7 +22,6 @@
 
 #include "RAJA/policy/hip/raja_hiperrchk.hpp"
```
MrBurmark, line 24:
```suggestion


```

src/basic/MAT_FUSED_MUL_ADD.hpp
```diff
@@ -0,0 +1,91 @@
...
+  Real_ptr A = m_A; 						                     \
+  Real_ptr B = m_B; 								     \
+  Real_ptr D = m_D; 						
+
+#define MAT_FUSED_MUL_ADD_BODY                                               	     \
+  Real_type dot = 0;                                                         	     \
+  for (Index_type k = 0; k < Ne; ++k) {                                      	     \
+    dot += A[row*Ne + k + ii*(Ne*Ne)] * B[k*Ne + col + ii*(Ne*Ne)];          	     \
+  }                                                                          	     \
+  D[row*Ne + col + ii*(Ne*Ne)] = dot;                                        	     \
```
MrBurmark, line 44: Is this supposed to be doing `D = A*B` or `d = A*B+C`?

src/basic/MAT_FUSED_MUL_ADD-Seq.cpp
```diff
@@ -0,0 +1,98 @@
...
+#if defined(RUN_RAJA_SEQ)
+  case Lambda_Seq: {
+
+    auto mat_fused_lam = [=](Index_type ii, Index_type row, Index_type col){
+        MAT_FUSED_MUL_ADD_BODY;
+        };
+
+    startTimer();
+    for (Index_type irep = 0; irep < run_reps; ++irep) {
+        for(Index_type ii = 0; ii != (N/(Ne*Ne)); ++ii){
```
MrBurmark, line 55: Is N/(Ne*Ne) the number of elements, should we make it a named quantity?

src/basic/MAT_FUSED_MUL_ADD.hpp
CRobeck: The mfma instructions are for computing D = A x B + C but we're assuming for this case C is zeros ignored. It's defined this way so we can expand future cases with a non-trivial C matrix.

MrBurmark: The name of the kernel is confusing then if we're not actually doing the ADD part.

src/apps/DEL_DOT_VEC_2D-Hip.cpp
```diff
@@ -122,7 +122,7 @@ void DEL_DOT_VEC_2D::runHipVariantImpl(VariantID vid)
 
       const size_t grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
 
-      hipLaunchKernelGGL((lambda_hip_forall<block_size, decltype(deldotvec2d_lambda)>),
+      hipLaunchKernelGGL((lambda_hip_forall_1D<block_size, decltype(deldotvec2d_lambda)>),
```
MrBurmark, line 125: Should this update be in another PR to keep this focused on the Kernel? I try to keep cuda and hip in sync.

CRobeck: I can revert it and/or move it to a new branch, it got pulled in from one of your review comments and I foresee adding a  lambda_hip_forall using 2D thread indexing and wanted to get in front of it.

MrBurmark: Let's not rename it here.

src/common/HipDataUtils.hpp
CRobeck: Well, currently we're only using it for testing for gfx908 and gfx90a supported operations so it should be OK but yes more generally the full name will give you xnack information as well. An example would be like:
gfx1010:sramecc-:xnack-
So if we want to use this function for that we should grab the entire string and then sub select it based on what we're going to query.

src/basic/MAT_FUSED_MUL_ADD-Hip.cpp
```diff
@@ -0,0 +1,304 @@
...
+  getHipDeviceData(m_B, B, N);                     \
+  getHipDeviceData(m_D, D, N);                     \
+  deallocHipDeviceData(A);                         \
+  deallocHipDeviceData(B);                         \
+  deallocHipDeviceData(D);						   
+
+__global__ void mat_fused_mul_add_builtin(const Real_ptr A, const Real_ptr B, Real_ptr D, const Index_type N){ 
+  constexpr Index_type Ne = 16;
+  const Index_type N_Elem = N/(Ne*Ne);
+  for(Index_type ii = 0; ii != N_Elem; ++ii){
```
MrBurmark, line 39: Do we want to parallelise over elements?

CRobeck: Since this is to mirror an FE mass matrix solve we need to assume that each element is independent and shares no common data (they could but this is the worst case).

PR #247, opened 2022-06-03
[Draft] Refactor GPU reductions and add unsafe atomic tunings

MrBurmark:
# Refactor GPU reductions and add unsafe atomic tunings

This adds a "unsafeAtomic" tuning of each of the kernels with a hip variant using atomics.
This also refactors gpu reductions so the implementation is not duplicated in each kernel with a reduction.
This also adds lambda variants of gpu reduction kernels.

- This PR is a refactoring and feature
  - refactors gpu reduction code to avoid duplication
  - Adds "unsafeAtomic" hip atomic tunings at the request of myself


rhornung67 (approved):
A lot of changes in this PR. I think I understand them all.

MrBurmark:
I kept finding things to change...
I have more in some more in mind for a future PR too.

src/algorithm/REDUCE_SUM-Hip.cpp
```diff
@@ -35,37 +35,23 @@ namespace algorithm
...
-    RAJA::atomicAdd<RAJA::hip_atomic>( dsum, psum[ 0 ] );
-  }
-#else // this doesn't work due to data races
-  if ( threadIdx.x == 0 ) {
-    *dsum += psum[ 0 ];
-  }
-#endif
+template < size_t block_size >
+__launch_bounds__(block_size)
+__global__ void reduce_sum_unsafe(Real_ptr x, Real_ptr dsum, Real_type sum_init,
```
CRobeck, line 51: If we start to add more tests with atomics we might not want to splitting these up into separate kernel for compile time concerns but should be OK for now

src/common/HipDataUtils.hpp
```diff
@@ -26,6 +26,47 @@
...
+inline std::string getHipArchName()
+{
+  int dev = -1;
+  hipDeviceProp_t devProp;
+  hipErrchk(hipGetDevice(&dev));
+  hipErrchk(hipGetDeviceProperties(&devProp, dev));
+  return devProp.gcnArchName;
+}
+
+#if defined(__gfx90a__)
```
CRobeck, line 41: Will be able to do away with __gfx90a__ arch check in the future

MrBurmark:
I'm closing this PR for a number of reasons. Its worth differentiating between safe and unsafe atomics as that is a temporary issue. The reducer implementation is difficult to put in a macro and not identically duplicated, so its not a huge gain to abstract it. The lambda variants are not necessary.

PR #260, opened 2022-08-24
HALOEXCHANGE_FUSED WorkGroup Dispatch Policies 

MrBurmark:
# Use new Dispatch policies in HALOEXCHANGE_FUSED.

This demonstrates the work group dispatch policies added to RAJA.

Based on changes in RAJA PR https://github.com/LLNL/RAJA/pull/1320.

- This PR is a feature
- It does the following:
  - Adds a point for dispatch policy comparison at the request of myself


robinson96 (approved):
LGTM

MrBurmark:
I've discovered one potential down side to abstracting the workgroup implementation. I can no longer use lambdas in the RAJA variants because the templated function has to support direct dispatch polices and indirect policies. The direct polices require fore-knowledge of the types used so I end up using functional classes instead of lambdas. Previously the indirect polices where in their own function so they used lambdas. I will be able to use `if constexpr` in c++17 to use lambdas however.

rhornung67 (approved):

PR #271, opened 2022-11-01
fix race condition in FIRST-MIN cuda/hip variant

CRobeck:
Fixes a race condition in the FIRST-MIN test. Currently, if you have more than one block, multiple threads will try to write to the output struct's memory in contention. In this PR we instead write out per block results and then finalize on the host.


CRobeck:
Still need to update Cuda variant

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,38 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
 
        FIRST_MIN_MINLOC_INIT;
 
-       MyMinLoc* dminloc;
-       cudaErrchk( cudaMalloc( (void**)&dminloc, sizeof(MyMinLoc) ) );
+       const size_t grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
+       MyMinLoc** dminloc;
```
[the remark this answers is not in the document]
CRobeck: Just seemed easier to do with pointers since we're passing them to kernels. But, yeah we could do it as fixed length arrays. We know the dims of the dminloc array.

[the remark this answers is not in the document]
CRobeck: @MrBurmark should be cleaned up now

rhornung67 (dismissed):
Thanks @CRobeck

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,31 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
 
+       const size_t grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
        MyMinLoc* dminloc;
-       cudaErrchk( cudaMalloc( (void**)&dminloc, sizeof(MyMinLoc) ) );
-       cudaErrchk( cudaMemcpy( dminloc, &mymin, sizeof(MyMinLoc),
-                               cudaMemcpyHostToDevice ) );
+       MyMinLoc mymin_block[grid_size]; //per-block min value
+
+       for (Index_type i=0;i<static_cast<Index_type>(grid_size);i++){
+	       memcpy(&(mymin_block[i]),&mymin, sizeof(MyMinLoc));
```
MrBurmark, line 88: Can we just use normal assignment here?

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,31 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
 
        FIRST_MIN_MINLOC_INIT;
 
+       const size_t grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
        MyMinLoc* dminloc;
-       cudaErrchk( cudaMalloc( (void**)&dminloc, sizeof(MyMinLoc) ) );
-       cudaErrchk( cudaMemcpy( dminloc, &mymin, sizeof(MyMinLoc),
-                               cudaMemcpyHostToDevice ) );
+       MyMinLoc mymin_block[grid_size]; //per-block min value
```
MrBurmark, line 85: I'd be happier with an allocated buffer than a runtime sized array.

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,31 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
                                                    iend );
-       cudaErrchk( cudaGetLastError() );
+       cudaErrchk( cudaMemcpy( &mymin_block, dminloc, grid_size * sizeof(MyMinLoc),
+                              cudaMemcpyDeviceToHost ) );       
 
-       cudaErrchk( cudaMemcpy( &mymin, dminloc, sizeof(MyMinLoc),
-                               cudaMemcpyDeviceToHost ) );
-       m_minloc = RAJA_MAX(m_minloc, mymin.loc);
+       for (Index_type i=0;i<static_cast<Index_type>(grid_size);i++){
+           m_minloc = RAJA_MAX(m_minloc, (mymin_block[i]).loc);
```
MrBurmark, line 102: Don't we want the min here?

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,31 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
+       }
```
MrBurmark, line 103: whitespace

src/lcals/FIRST_MIN-Cuda.cpp
rhornung67: Good catch. I missed that.

CRobeck: Then I think we need to update the other variants right? Or am I interpreting what the algorithm is incorrectly?

CRobeck: OK. I think we should be consistent now.

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,35 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
+                              cudaMemcpyDeviceToHost ) );       
+
+       for (Index_type i=0;i<static_cast<Index_type>(grid_size);i++){
+           if( (mymin_block[i]).val < mymin.val)
+                   mymin = mymin_block[i];
+       }
 
-       cudaErrchk( cudaMemcpy( &mymin, dminloc, sizeof(MyMinLoc),
-                               cudaMemcpyDeviceToHost ) );
        m_minloc = RAJA_MAX(m_minloc, mymin.loc);
```
MrBurmark, line 108: @rhornung67 shouldn't this just be an assignment? What if it was already right and the result was smaller and wrong, then it would not overwrite the existing value and we'd think it was right when it was actually wrong.

MrBurmark: I wouldn't change it here, but I'm not sure this is quite right.

src/lcals/FIRST_MIN-Hip.cpp
```diff
@@ -80,22 +80,32 @@ void FIRST_MIN::runHipVariantImpl(VariantID vid)
...
+                              hipMemcpyDeviceToHost ) );       
+
+       for (Index_type i=0;i<static_cast<Index_type>(grid_size);i++){
+	   if( (mymin_block[i]).val < mymin.val) 
+		   mymin = mymin_block[i];
+       }	   
+
        m_minloc = RAJA_MAX(m_minloc, mymin.loc);
 
+       hipErrchk( hipGetLastError() );
```
MrBurmark, line 108: extra

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,24 +80,35 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
 
+       cudaErrchk( cudaGetLastError() );
```
MrBurmark, line 110: extra

src/lcals/FIRST_MIN-Cuda.cpp
MrBurmark: In fact, would it be better if this was pinned memory in the first place so we could avoid the memcpy?

src/lcals/FIRST_MIN-Cuda.cpp
rhornung67: I think it can be an assignment since MyMinLoc is a POD type: https://github.com/LLNL/RAJAPerf/blob/develop/src/lcals/FIRST_MIN.hpp#L41

CRobeck: Updated the comparison method

src/lcals/FIRST_MIN-Cuda.cpp
rhornung67: @MrBurmark would that be a fair comparison with the RAJA variants?

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -27,6 +27,8 @@ namespace lcals
 #define FIRST_MIN_DATA_TEARDOWN_CUDA \
   deallocCudaDeviceData(x);
 
+FIRST_MIN_MINLOC_COMPARE;
```
rhornung67, line 30: This line should be removed. It causes a multiply defined symbol linker error.

src/lcals/FIRST_MIN-Hip.cpp
```diff
@@ -27,6 +27,8 @@ namespace lcals
 #define FIRST_MIN_DATA_TEARDOWN_HIP \
   deallocHipDeviceData(x);
 
+FIRST_MIN_MINLOC_COMPARE;
```
rhornung67, line 30: This line should be removed. It causes a multiply defined symbol linker error.

CRobeck: Hmm. Needs to be forward declared somewhere or else we get a hard compile error.

rhornung67: You're right. It works with CUDA because we typically enable OpenMP and CUDA, but we haven't been turning on OpenMP with HIP builds. It's forward declared in FIRST_MIN-OMP.cpp. I recently fixed the RAJA CMake stuff to support OpenMP with HIP also.

I think the right thing to do is to forward declare in FIRST_MIN-Cuda.cpp and FIRST_MIN-Hip.cpp like you had it and guard the declarations in those spots with #if !defined(RAJA_ENABLE_OPENMP). Does that make sense?

rhornung67: Maybe better to just forward declare once in FIRST_MIN.cpp, which will get compiled in any case.

rhornung67: My ideas here don't work for all cases. I will push a change to fix the issue shortly.

src/lcals/FIRST_MIN-Cuda.cpp
CRobeck: Right, I worry pinned memory here will make Cuda/HIP comparisons challenging.

src/lcals/FIRST_MIN-Cuda.cpp
rhornung67: > @rhornung67 shouldn't this just be an assignment? What if it was already right and the result was smaller and wrong, then it would not overwrite the existing value and we'd think it was right when it was actually wrong.

@MrBurmark I mis-interpreted your question when you asked it originally. Now, I see what you were asking. Yes, that can be an assignment and probably should be to make the code more clear. I'll push a change.

rhornung67:
@CRobeck this should be good now. We'll merge it after @MrBurmark takes another look.

rhornung67 (approved):

CRobeck:
> @CRobeck this should be good now. We'll merge it after @MrBurmark takes another look.

Thanks for the assist

rhornung67:
Ping @MrBurmark Any concerns or suggestions on this?

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -80,21 +81,36 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
+       for (Index_type i = 0; i < static_cast<Index_type>(grid_size); i++) {
+         mymin_block[i] = mymin;
+       }
+
        MyMinLoc* dminloc;
-       cudaErrchk( cudaMalloc( (void**)&dminloc, sizeof(MyMinLoc) ) );
-       cudaErrchk( cudaMemcpy( dminloc, &mymin, sizeof(MyMinLoc),
-                               cudaMemcpyHostToDevice ) );
+       cudaErrchk( cudaMalloc( (void**)&dminloc, 
+                               grid_size * sizeof(MyMinLoc) ) );
```
MrBurmark, line 93: Can we move some of this out of the inner irep loop?

src/lcals/FIRST_MIN-Cuda.cpp
MrBurmark: Pinned will be faster, so it will be a better baseline to compare against.

src/lcals/FIRST_MIN-Cuda.cpp
MrBurmark: This looks good.

src/lcals/FIRST_MIN-Cuda.cpp
CRobeck: Moved some init functions outside the run rep loop. Let me know if that was what you were thinking? I'm debating about what is a fair comparisons with other variants.

rhornung67: I'm working on a few changes that I think Jason had in mind. I will push in a bit. 

It looks there is a memory leak here since the device malloc of dminloc is inside the irep loop, but the free is outside of the loop. I will address in my changes.

rhornung67:
@CRobeck I believe my last change set resolves the orphaned device memory issue and aligns the base CUDA and HIP variants with their RAJA counterparts.  @MrBurmark please take a look and see if you agree.

CRobeck:
Looks good. I caught the memory leak right after I pushed it.

src/lcals/FIRST_MIN-Cuda.cpp
```diff
@@ -75,32 +76,45 @@ void FIRST_MIN::runCudaVariantImpl(VariantID vid)
...
-                               cudaMemcpyDeviceToHost ) );
-       m_minloc = RAJA_MAX(m_minloc, mymin.loc);
+      FIRST_MIN_MINLOC_INIT;
 
-       cudaErrchk( cudaFree( dminloc ) );
+      for (Index_type i = 0; i < static_cast<Index_type>(grid_size); i++) {
+        mymin_block[i] = mymin;
+      }
+      cudaErrchk( cudaMemcpy( dminloc, mymin_block, grid_size*sizeof(MyMinLoc),
+                              cudaMemcpyHostToDevice ) );
```
MrBurmark, line 95: We can get rid of this, we overwrite the memory each time.

src/lcals/FIRST_MIN-Cuda.cpp
MrBurmark: Its fair in my opinion to allocate outside the loop as RAJA uses a memory pool instead of allocating new memory each time.

src/lcals/FIRST_MIN-Cuda.cpp
rhornung67: All of it, the for-loop and the memcpy?  Isn't that representative of the RAJA CUDA variant, here?

https://github.com/LLNL/RAJAPerf/blob/develop/src/lcals/FIRST_MIN-Cuda.cpp#L113

MrBurmark: Nope, the RAJA cuda/hip variants don't do any any of this kind of setting memory or copying memory.

rhornung67: Because we account for the initial value for the reduction during finalization?

CRobeck: We originally had a RAJA_MAX here because mymin could have an initial value so we check against that:

https://github.com/LLNL/RAJAPerf/blob/develop/src/lcals/FIRST_MIN.hpp#L51

Since we got rid those maxes having an init value is redundant. Are we OK not having an init value here?

CRobeck: The RAJA does have an init value that gets captured by the lambda

rhornung67: @CRobeck after conferring with @MrBurmark, we agreed that this follows the memory setting/motion behavior of the RAJA variants the best. If this were a functionality test, not a performance comparison we would do it differently.

CRobeck: Sounds good.

MrBurmark (approved):

PR #297, opened 2023-01-21
[WIP] Add unified memory space HIP variants

CRobeck:
This PR is the first in a series to add/allow for more fine tuning within a unified memory space with the perf suite HIP kernels.

### DIFF
--- a/src/algorithm/MEMCPY-Hip.cpp
+++ b/src/algorithm/MEMCPY-Hip.cpp
@@ -21,12 +21,18 @@ namespace rajaperf
 namespace algorithm
 {
 
-#define MEMCPY_DATA_SETUP_HIP \
-  allocAndInitHipDeviceData(x, m_x, iend); \
-  allocAndInitHipDeviceData(y, m_y, iend);
+#define MEMCPY_ALLOC_HIP_DATA \
+  allocHipDeviceData(x, iend); \
+  allocHipDeviceData(y, iend);
 
-#define MEMCPY_DATA_TEARDOWN_HIP \
-  getHipDeviceData(m_y, y, iend); \
+#define MEMCPY_INIT_HIP_DATA \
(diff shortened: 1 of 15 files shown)


MrBurmark:
@CRobeck What is the final intention, to allow different ways to allocate memory? If so I started looking at that with #250. If not I'm not sure what capability this PR is trying to add.
Those macros are not in the timed section of the code if you're worried about performance. I'm not sure how useful it is to break one macro into two if they're always just used back to back anyway.

CRobeck:
Right now the memory allocations and the H2D memcpys are bundled into the same macro. I'd like to make a tuning for the unified memory space in which we don't need to call the memcpys. Separating them out so I could omit the memcpy seemed like the least cluttered approach.

MrBurmark:
The allocation+init/copy call is made in basically every kernel variant file so reducing the number of function calls there is probably best even if it increases the amount of functions in the utils.
So I'm in favor of something like having a generic allocate+init/copy instead of splitting the calls as that keeps the code as simple as possible in the kernel files. It probably makes sense to make the high level allocate+init/copy function in KernelBase as it has access to the run parameters  like here https://github.com/LLNL/RAJAPerf/pull/296/commits/c44990e0127eb5efed086f86cc566507a7a99ca5#diff-f1b7ede2b59bf264efbe43bc92349deb28c8f49dc87efa11b4882ab6805beb24R174 in #296 .

CRobeck:
So you're thinking move
https://github.com/LLNL/RAJAPerf/blob/develop/src/common/HipDataUtils.hpp#L141
into KernelBase and then get rid of the alloc/init macros in each variant? That seems fine but we still need a way of decoupling the alloc from the memcpy. Also,iIsn't allocAndInitHipDeviceData already a "generic allocate+init/copy"?

PR #299, opened 2023-02-08
Fix GPU FIRST_MIN implementations

MrBurmark:
# Fix FIRST_MIN Base GPU variants

Previously they used the output array before it was initialized, now pass in an init value.

- This PR is a bugfix
  - Fixes something I noticed


rhornung67 (approved):
Good catch!

CRobeck:
Are there any cases in which we would want different init values for each block? i.e. init the dminloc array? If not then LGTM.

CRobeck (approved):

artv3 (approved):

MrBurmark:
I don't think there is a reason to have different init values per block for this kernel.

PR #320, opened 2023-03-27
removed Apps_COUPLE encumbrance wrt embedded code to exclude this ke…

jonesholger:
Remove Apps_Couple and place it into TODO directory

Apps_COUPLE had explicit encumbrance  to exclude kernel since it was still WIP and needed lightweight device side complex math

This kernel hindered the generation of a RAJAPerf-minimal implementation to add and test a single kernel

This begs the question of adding infrastructure to RAJAPerf to add specific TODO kernels into the framework for testing with:
1: cmake -DRAJAPERF_ADD_TODO
2: add cmake macros into src/CMakeLists to add_subdir_list
3: add sed marker text and ifdef guards into RAJAPerfSuite.hpp/cpp for TODO Kernel

This infrastructure could also support RAJAPerf tutorial to add arbitrary kernel


rhornung67 (approved):
Thanks, @jonesholger

PR #342, opened 2023-07-05
New RAJA GPU Indexing

MrBurmark:
# Use new RAJA GPU indexing policies

This gives compile time knowledge of the block sizes in some kernels. This brings the RAJA variants closer in line with the base variants.
The base variants can still take a advantage of the knowledge that the ranges start at 0.

- This PR is a feature
- It does the following:
  - Adds an example of using the new GPU indexing policies at the request of @rhornung67


MrBurmark:
Updated this so we can get new performance results.

MrBurmark:
@rhornung67 @artv3  Can I get a review?

rhornung67 (approved):
Did you do all the kernels that use RAJA::kernel and RAJA::launch? I was going to work through that, but thank you for doing it @MrBurmark

artv3 (approved):

MrBurmark:
I think I did all the kernels.

Issue #381, opened 2023-10-24
Add kernels broken up over multiple loops


MrBurmark:
Add version of some kernels that break the iteration space into multiple subdomains and run a loop over each.
The intention is to compare performance when using a single loop over all the data or using multiple loops to cover multiple subsections or tiles of the data.
Kernels to look at first:
- Stream_TRIAD
- Algorithm_REDUCE_SUM
- Basic_INDEXLIST_3LOOP

PR #382, opened 2023-10-25
Add TRIAD_PARTED kernel

MrBurmark:
# Add TRIAD_PARTED kernel

This does the same thing as TRIAD but breaks it into multiple for loops over the data instead of a single for loop over the data.
The idea is to look at ways of improving the performance of kernels when running the same kernel over different data like when running over subdomains or tiles.

- This PR is a feature
- It does the following:
  - Adds parted kernel at the request of me in #381


MrBurmark:
I'm considering doing this for the kernels listed in #381.
I may add a folder to partition this into a single place.

rhornung67 (approved):

PR #383, opened 2023-10-25
Fix scripts

MrBurmark:
# Fix our build scripts

Enable cuda relocatable device code in the scripts to ensure kernel fusion works.
Make the scripts more consistent.

Note that we should also add rdc to our hip build and we should add rdc to our CI builds if its not there already.

- This PR is a bugfix
  - Fixes cuda crash at runtime


rhornung67 (approved):
Thanks @MrBurmark Should we also make similar changes in the RAJA build scripts?

PR #389, opened 2023-11-13
[Woptim] alloc tuning

adrienbernede:
### Summary

Investigate potential of speed up in GitLab: reduce allocation duration, fill up nodes, group allocations...

### DIFF
--- a/.gitlab/custom-jobs-and-variables.yml
+++ b/.gitlab/custom-jobs-and-variables.yml
@@ -15,9 +15,9 @@ variables:
 
 # Ruby
 # Arguments for top level allocation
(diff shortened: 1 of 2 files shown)


adrienbernede:
The previous allocation were well beyond necessary. The new ones are like 20% above what is usually necessary, and 1 extra minute is generally granted when a job exceeds it’s allocated time.

Also, what I am observing is that the variability often comes from the wait time for the allocation. Once the allocation is granted, performance is rather stable (we are allocating nodes exclusively).

I could be wrong, if you think we should not try those values and take more margin, let me know.

PR #419, opened 2024-02-17
Add MULTI_REDUCE and HISTOGRAM Kernels

MrBurmark:
# Add MULTI_REDUCE and HISTOGRAM Kernels

Add the MULTI_REDUCE and HISTOGRAM kernels which look at the performance for reducing into a runtime number of bins. In MULTI_REDUCE each iterate contributes to a single bin in the output like a histogram. HISTOGRAM allows us to have a comparison point with GPU performance libraries like cub and rocprim.

Which iterate maps to which bin is currently random, but this is not really a realistic use case. More realistic use cases are contiguous chunks of the domain mapping to the same bin, or even the entire domain mapping to a single bin.
The CPU parallel implementations are not optimized and simple use atomics into an array with no replication.

- This PR is a feature
  - Adds Multi-Reduce example at the request of me


artv3 (approved):

src/algorithm/HISTOGRAM-Cuda.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+    // Allocate temporary storage
+    unsigned char* temp_storage;
+    allocData(DataSpace::CudaDevice, temp_storage, temp_storage_bytes);
+    d_temp_storage = temp_storage;
+
+    startTimer();
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      // Run
```
artv3, line 81: go kernel go!

MrBurmark: Let's remove this pointless comment.

MrBurmark: Never-mind, this is with cub/rocprim stuff that is a bit confusing about when it actually runs.

MrBurmark:
Interestingly, before rocm 6.1 things like threadIdx.x were not integral types, which is why the rocm test is failing here.

MrBurmark:
> Interestingly, before rocm 6.1 things like threadIdx.x were not integral types, which is why the rocm test is failing here. We should check that RAJA won't force people to rocm 6.1.

I cast the uses of hip dimensions like threadIdx.x when used with RAJA::power_of_2_mod in RAJAPerf and confirmed that RAJA doesn't run into this particular stumbling block.

src/algorithm/HISTOGRAM-Cuda.cpp
```diff
@@ -0,0 +1,356 @@
...
+        seq_for(cuda_atomic_global_replications_type{}, [&](auto global_replication) {
+
+          if (run_params.numValidAtomicReplication() == 0u ||
+              run_params.validAtomicReplication(global_replication)) {
+
+            seq_for(cuda_atomic_shared_replications_type{}, [&](auto shared_replication) {
+
+              addVariantTuningName(vid, "atomic_shared("+std::to_string(shared_replication)+
+                                        ")_global("+std::to_string(global_replication)+
+                                        ")_"+std::to_string(block_size));
```
MrBurmark, line 337: I left the global and shared replication tuning parameters in here so we could try out a sweep over them later to tune for new HW, but it does end up leaving the tunings with fairly long names. It also causes me to have different default tuning params for cuda and hip. Finally to do this I am writing the RAJA multi_reduce policy from a low level and not using the high level policy so its possible for the tuning here and the tuning in RAJA to get out of sync.

MrBurmark: I also did not add the with_reduce policy vs normal policy option that the reduce kernels have, though I expect the with_reduce policies to be faster as they use fewer global atomics.

MrBurmark: I added the occupancy vs normal policy option

PR #453, opened 2024-06-11
Use GitLab container registry as a Spack build cache

adrienbernede:
In this PR, we set the GitLab CI to push Spack built binaries (Umpire dependencies) to the builtin OCI container registry.

This required an update of Spack, RADIUSS Spack Configs and Uberenv.


adrienbernede:
@rhornung67 this should go in before other CI changes to be on par with RAJA CI.

rhornung67 (approved):

PR #500, opened 2025-02-03
Testing direct vs loop stride 

artv3:
# Summary

This test modifies an existing kernel to use direct threading.

### DIFF
--- a/src/apps/FEM_MACROS.hpp
+++ b/src/apps/FEM_MACROS.hpp
@@ -26,9 +26,7 @@
 
 #if defined(RAJA_ENABLE_CUDA) || defined(RAJA_ENABLE_HIP)
 #define GPU_FOREACH_THREAD_DIRECT(i, k, N)  \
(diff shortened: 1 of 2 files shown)


src/apps/FEM_MACROS.hpp
```diff
@@ -24,6 +24,11 @@
   for (int i = threadIdx.k; i < N; i += blockDim.k)
 #endif
 
+#if defined(RAJA_ENABLE_CUDA) || defined(RAJA_ENABLE_HIP)
+#define GPU_FOREACH_THREAD_DIRECT(i, k, N)  \
+  if(int i = threadIdx.k; i <  N)
```
khuck-AMD, line 29: I am confused about this line, @artv3 - did you intend for this to be an `if` statement?

khuck-AMD: never mind, I seem to understand what the code is doing now

PR #505, opened 2025-03-29
Fix optimization level on a few Caliper LC build scripts

berg-michael:
# Summary

- This PR is a bugfix
  - Fixes a few LC build scripts which appear to configure the build without optimization and with debugging info. The host configs generate conflicting flags if configured in release mode. 
  - It seems that those settings (in the host configs) take precedence over these flags because they appear later in the compilation command. Nevertheless, I don't think the build scripts should have these flags.


artv3 (approved):
Yes, I think is is a good idea.

PR #514, opened 2025-05-22
Improve how copies are handled

MrBurmark:
# Improve copies

This does a number of things to improve how copies are handled.
Add synchronization to copyData to ensure those copies complete. This fixes a potential issue around host->device and device->device copies with cuda and hip.
Use the programming model memcpy implementation instead of copyData in some sections of the code. This fixes a potential issue where a sequential memcpy implementation was used for host to host copies with cuda and hip that could lead to a host-device race condition.
Make a combined scopedMoveData and allocData API called allocDataForInit, along with variants that initialize memory. This generally reduces the number of allocations and copies used by allocating directly into the memory space used during initialization. This simplifies usage in setup routines, but makes allocate and copy patterns elsewhere more explicit.

- This PR is a refactoring, bugfix
- It does the following:
  - Modifies/refactors scopedMoveData to combine it with allocation and init and give it a better name.
  - Fixes a potential host/device race condition issue

### DIFF
--- a/src/common/KernelBase.hpp
+++ b/src/common/KernelBase.hpp
@@ -419,6 +419,33 @@ public:
   }
 
   template <typename T>
+  rajaperf::AutoDataMover<T> allocDataForSeqInit(T*& ptr, Size_type len, VariantID vid)
+  {
+    DataSpace ds = getDataSpace(vid);
+    DataSpace hds = rajaperf::hostCopyDataSpace(ds);
+    rajaperf::allocData(hds, ptr, len, getDataAlignment());
+    return {ds, hds, ptr, len, getDataAlignment()};
+  }
+
+  template <typename T>
+  rajaperf::AutoDataMover<T> allocAndInitDataForSeqInit(T*& ptr, Size_type len, VariantID vid)
+  {
+    DataSpace ds = getDataSpace(vid);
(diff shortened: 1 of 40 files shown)


src/common/KernelBase.hpp
```diff
@@ -418,6 +418,33 @@ class KernelBase
...
+  }
+
+  template <typename T>
+  rajaperf::AutoDataMover<T> allocAndInitDataConstForInit(T*& ptr, Size_type len, T val, VariantID vid)
+  {
+    DataSpace ds = getDataSpace(vid);
+    DataSpace hds = rajaperf::hostCopyDataSpace(ds);
+    rajaperf::allocAndInitDataConst(hds, ptr, len, getDataAlignment(), val);
+    return {ds, hds, ptr, len, getDataAlignment()};
+  }
```
rchen20, line 446: Are these functions always allocating on the host, and only during initialization? Is that what `ForInit` implies?

rchen20 (approved):
Looks fine overall. Just one question for clarity.

src/common/KernelBase.hpp
MrBurmark: Yep, that's the intention. They should not be used during a timed region as they may have less than ideal performance. Sounds like I should document things better.

PR #527, opened 2025-07-03
Caliper - Use Native Type Instead of Casting to Fix Precision Issue

michaelmckinsey1:
# Summary

- This PR is a bugfix
  - Updates the attributes in `src/common/KernelBase.cpp` to use `Index_type` instead of casting to double, which loses precision after string conversion in caliper https://github.com/LLNL/Caliper/blob/cd94bc43a84f28dce61f4795a280067dbcc79c3f/src/common/Variant.cpp#L83 (during testing this is type `long`, so overflow won't happen for large bytes values). This would result in a value such as `1.34758e+08` evaluating to `134758000` instead of true value `134758042`.
  - Tested up to problem size of `536870912` (bytes > 2B to check int overflow)

### DIFF
--- a/src/common/KernelBase.cpp
+++ b/src/common/KernelBase.cpp
@@ -574,36 +574,29 @@ void KernelBase::print(std::ostream& os) const
 void KernelBase::doOnceCaliMetaBegin(VariantID vid, size_t tune_idx)
 {
   if(doCaliMetaOnce[vid].at(tune_idx)) {
-    // attributes are class variables initialized in ctor
-    int index_type_size = sizeof(Index_type);
+    // Set values for Index_type.
+    // Some of these may overflow if using "cali_set_int"
+    auto cali_set_helper = [](cali_id_t const& attr, Index_type val) {
+      cali_set(attr, &val, sizeof(Index_type));
+    };
+    cali_set_helper(ProblemSize_attr, getActualProblemSize());
+    cali_set_helper(Reps_attr, getRunReps());
+    cali_set_helper(Iters_Rep_attr, getItsPerRep());
+    cali_set_helper(Kernels_Rep_attr, getKernelsPerRep());
+    cali_set_helper(Bytes_Rep_attr, getBytesPerRep());
+    cali_set_helper(Bytes_Read_Rep_attr, getBytesReadPerRep());
+    cali_set_helper(Bytes_Written_Rep_attr, getBytesWrittenPerRep());
+    cali_set_helper(Bytes_AtomicModifyWritten_Rep_attr, getBytesAtomicModifyWrittenPerRep());
+    cali_set_helper(Flops_Rep_attr, getFLOPsPerRep());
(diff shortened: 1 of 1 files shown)


src/common/KernelBase.cpp
```diff
@@ -575,16 +575,30 @@ void KernelBase::doOnceCaliMetaBegin(VariantID vid, size_t tune_idx)
 {
   if(doCaliMetaOnce[vid].at(tune_idx)) {
     // attributes are class variables initialized in ctor
-    cali_set_double(ProblemSize_attr,(double)getActualProblemSize());
```
MrBurmark, line 578: Should these use whatever int type caliper uses instead of the RAJAPerf Index_type?

michaelmckinsey1: `cali_set_int` is 32-bit int which will overflow the bytes metrics for larger problem sizes. Ideally we would use a `cali_set_int64` function (doesn't exist), but `cali_set` is capable of storing the value of `Index_type`.

MrBurmark:
@michaelmckinsey1 is this still a draft?

michaelmckinsey1:
> @michaelmckinsey1 is this still a draft?

It is ready

src/common/KernelBase.cpp
```diff
@@ -575,16 +575,30 @@ void KernelBase::doOnceCaliMetaBegin(VariantID vid, size_t tune_idx)
...
-    cali_set_double(Reps_attr,(double)getRunReps());
-    cali_set_double(Iters_Rep_attr,(double)getItsPerRep());
-    cali_set_double(Kernels_Rep_attr,(double)getKernelsPerRep());
-    cali_set_double(Bytes_Rep_attr,(double)getBytesPerRep());
-    cali_set_double(Bytes_Read_Rep_attr,(double)getBytesReadPerRep());
-    cali_set_double(Bytes_Written_Rep_attr,(double)getBytesWrittenPerRep());
-    cali_set_double(Bytes_AtomicModifyWritten_Rep_attr,(double)getBytesAtomicModifyWrittenPerRep());
-    cali_set_double(Flops_Rep_attr,(double)getFLOPsPerRep());
-    cali_set_double(BlockSize_attr, getBlockSize());
+    int index_type_size = sizeof(Index_type);
```
MrBurmark: Would it make sense to make a little helper lambda?
```cpp
auto cali_set_helper = [](Attr const& attr, Index_type val) {
  cali_set(attr, &val, sizeof(Index_type));
};

cali_set_helper(Reps_attr, getRunReps());
...
```

michaelmckinsey1: Yes this does make sense, [a3ba1bb](https://github.com/LLNL/RAJAPerf/pull/527/commits/a3ba1bb80087d99fa2321c61f297ab5f2705c299). Thank you.

MrBurmark (approved):

PR #535, opened 2025-08-13
TESTING only: enforce larger time span for kernel execution

amroakmal:
# Summary

- This PR is a (refactoring, bugfix, feature, something else)
  - Modifies/refactors (class or method) (how?)
  - Fixes (issue number(s))

### DIFF
--- a/src/stream/TRIAD.hpp
+++ b/src/stream/TRIAD.hpp
@@ -26,7 +26,7 @@
 #define TRIAD_BODY  \
   a[i] = b[i] + alpha * c[i] ;
 
-constexpr int extra_kernel_reps = 5;
+constexpr int extra_kernel_reps = 20;
 
 #include "common/KernelBase.hpp"
 


src/stream/TRIAD-Hip.cpp
```diff
@@ -82,18 +84,26 @@ void TRIAD::runHipVariantImpl(VariantID vid)
...
-        RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i) {
-        TRIAD_BODY;
-      });
+      for (RepIndex_type extra_rep = 0; extra_rep < 5; ++extra_rep) {
+          getCout() << "\n  TRIAD : Unknown Hip variant id = " << vid << std::endl;
+          temp_count += extra_rep * (extra_rep & 1);
+          
+          RAJA::forall< RAJA::hip_exec<block_size, true /*async*/> >( res,
+            RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i) {
+              TRIAD_BODY;
```
MrBurmark: I think you want the extra rep loop around TRIAD_BODY

src/stream/TRIAD-Hip.cpp
```diff
@@ -82,18 +82,28 @@ void TRIAD::runHipVariantImpl(VariantID vid)
...
+            temp_count += extra_rep * (extra_rep & 1);
+            TRIAD_BODY;
+          }
+          
+      });
     }
+
     stopTimer();
 
+    getCout() << "\n Dummy print preventing compiler optimization: " << temp_count << std::endl;
```
MrBurmark: you can get rid of this.

src/stream/TRIAD-Hip.cpp
```diff
@@ -82,18 +82,28 @@ void TRIAD::runHipVariantImpl(VariantID vid)
     stopTimer();
 
   } else if ( vid == RAJA_HIP ) {
+    
+    RepIndex_type temp_count = 0;
```
MrBurmark: remove this

src/stream/TRIAD-Hip.cpp
```diff
@@ -82,18 +82,28 @@ void TRIAD::runHipVariantImpl(VariantID vid)
...
     startTimer();
+    
     for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
 
       RAJA::forall< RAJA::hip_exec<block_size, true /*async*/> >( res,
         RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i) {
-        TRIAD_BODY;
-      });
 
+          for (RepIndex_type extra_rep = 0; extra_rep < 5; ++extra_rep) {
```
MrBurmark: make a var for 5

MrBurmark: so its easier to change

src/stream/TRIAD-Hip.cpp
```diff
@@ -82,18 +82,28 @@ void TRIAD::runHipVariantImpl(VariantID vid)
...
+            temp_count += extra_rep * (extra_rep & 1);
```
MrBurmark: We can try adding EMPTY_BODY (https://github.com/LLNL/RAJAPerf/blob/cda42470851fff2b7c8e6a9b5b11ab83f33a5a07/src/basic/EMPTY.hpp#L29) if the compiler optimizes out the loop.

src/stream/TRIAD-Hip.cpp
amroakmal: Done

src/stream/TRIAD-Hip.cpp
amroakmal: Done

src/stream/TRIAD-Hip.cpp
amroakmal: So this in turns leads to removing "temp_count += extra_rep * (extra_rep & 1);"
I removed it, no more extra operations are done in the loop except TRIAD_BODY

src/stream/TRIAD.hpp
```diff
@@ -26,6 +26,7 @@
 #define TRIAD_BODY  \
   a[i] = b[i] + alpha * c[i] ;
 
+#define ENFORCED_KERNEL_REPETITION 5
```
MrBurmark: Make this a runtime variable

MrBurmark: like num_reps

src/stream/TRIAD-Hip.cpp
```diff
@@ -82,16 +82,21 @@ void TRIAD::runHipVariantImpl(VariantID vid)
     stopTimer();
 
   } else if ( vid == RAJA_HIP ) {
-
+
```
MrBurmark: spacing

src/stream/TRIAD.hpp
amroakmal: Done

src/stream/TRIAD.hpp
```diff
@@ -26,6 +26,7 @@
 #define TRIAD_BODY  \
   a[i] = b[i] + alpha * c[i] ;
 
+constexpr int extra_kernel_rep_count = 5;
```
MrBurmark:
```suggestion
constexpr int extra_kernel_reps = 5;
```

MrBurmark: to match num_reps

src/stream/TRIAD-Hip.cpp
amroakmal: Done

src/stream/TRIAD.hpp
amroakmal: Done

src/basic/MAT_MAT_SHARED-Hip.cpp
```diff
@@ -86,92 +86,93 @@ void MAT_MAT_SHARED::runHipVariantImpl(VariantID vid)
     for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
 
       auto mat_mat_shared_lambda = [=] __device__() {
+        for (RepIndex_type extra_rep = 0; extra_rep < extra_kernel_reps; ++extra_rep) {
```
MrBurmark: When adding something like this for debugging, its ok to not indent the code in the loop. That will make it easier to see that the only thing done was adding a loop around this part of the code.

src/basic/MAT_MAT_SHARED.hpp
```diff
@@ -119,6 +119,8 @@ constexpr rajaperf::Index_type TL_SZ = 16;
   if (Row < N && Col < N)                                                      \
     C[Col + N * Row] = Cs[ty][tx];
 
+constexpr int extra_kernel_reps = 5;
```
MrBurmark: You'll want to make a local variable of this and put it into the DATA_SETUP macro

MrBurmark: Its probably best to add this as a member variable of KernelBase so the compiler doesn't know how many extra_kernel_reps at compile time.

src/basic/MAT_MAT_SHARED-Hip.cpp
amroakmal: Okay, thanks for this! Done :)

src/basic/MAT_MAT_SHARED.hpp
amroakmal: Sure, for now I added as you advised in the first comment, and will also try to do what you advised in this comment. Thanks

MrBurmark:
I expect this approach to cause the tests to fail, so don't worry too much about getting the tests to pass.

PR #529, opened 2025-07-16
[WIP] Caliper runs in CI

adrienbernede:
# Summary

This PR acts as a demo for https://github.com/LLNL/radiuss-shared-ci/pull/46.

It leverages dedicated performance pipeline to run caliper analysis on selected machines (here dane and tuolumne).

TODO:
- [ ] Turn corona back ON (.gitlab-ci.yml): was turned off because of a failure preventing to go all the way to perf tests.
- [x] Merge corresponding branch in radiuss-shared-ci first (https://github.com/LLNL/radiuss-shared-ci/pull/63)
- [ ] Be sure to test with bradiuss workflow token

PR #545, opened 2025-09-15
Caliper benchmarks in CI (part 2)

tdrwenski:
Additional changes to https://github.com/LLNL/RAJAPerf/pull/529 (to merge into that branch):

- Add caliper to json conversion script
- Add GitHub workflow to handle benchmarks and create dashboard and alerts
- Update build and test script to run raja_perf with specified kernels/ variants
- Update performance jobs

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,8 +15,11 @@ jobs:
   check-performance:
     if: github.ref != 'refs/heads/develop'
     permissions:
+      # Needed to write PR check status
       checks: write
-      contents: write # TODO for test, later make it read
+      # Need write if you turn on comments
+      pull-requests: read
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
@@ -35,7 +38,7 @@ jobs:
       - name: Decode benchmark data
         run: echo "${{ github.event.inputs.benchmark_data }}" | base64 -d > benchmark.json
 
-      - name: Store benchmark result
+      - name: Compare benchmark result
         id: benchmark
         continue-on-error: true
         uses: benchmark-action/github-action-benchmark@v1
@@ -44,11 +47,12 @@ jobs:
           tool: 'customSmallerIsBetter'
           output-file-path: benchmark.json
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          auto-push: true # TODO for testing, later make it false
+          auto-push: false
+          save-data-file: false
           gh-pages-branch: gh-pages
           benchmark-data-dir-path: dev/bench
           comment-always: false
-          comment-on-alert: true
+          comment-on-alert: false
           summary-always: true
           fail-on-alert: true
           alert-threshold: '120%'
@@ -74,8 +78,9 @@ jobs:
         run: exit 1
 
   store-and-visualize-on-develop:
-    if: github.ref == 'refs/heads/develop'
+    # if: github.ref == 'refs/heads/develop' # TODO uncomment before merging
     permissions:
+      # Needed to push to gh-pages branch
       contents: write
     runs-on: ubuntu-latest
     steps:


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Store benchmark result
+        id: benchmark
+        continue-on-error: true
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: ${{ github.event.inputs.benchmark_name }}
+          tool: 'customSmallerIsBetter'
+          output-file-path: benchmark.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true # TODO remove later, for testing
```
adrienbernede: "Remove later" like now? Or in a subsequent PR?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,76 @@
...
+        required: true
+      benchmark_name:
+        description: 'Name for benchmark group on the dashboard'
+        required: false
+        default: 'RAJAPerf benchmarks'
+
+permissions:
+  checks: write
+  contents: write
+  pull-requests: write
```
adrienbernede: Double checking: are all those permissions required?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,76 @@
...
+          comment-always: false
+          comment-on-alert: true
+          summary-always: true
+          fail-on-alert: true
+          alert-threshold: '120%'
+          max-items-in-chart: 100
+
+      - name: Create Check for PRs
+        if: steps.findpr.outputs.pr_number != ''
+        uses: LouisBrunner/checks-action@v2.0.0
```
adrienbernede: Maybe use the commit hash to specify the version... a tag can be changed, and I’m a bit worried by the low maintenance activity on this repo.

scripts/gitlab/build_and_test.sh
```diff
@@ -30,6 +30,11 @@ spack_debug=${SPACK_DEBUG:-false}
 debug_mode=${DEBUG_MODE:-false}
 push_to_registry=${PUSH_TO_REGISTRY:-true}
 perf_artifact_dir=${PERF_ARTIFACT_DIR:-""}
+performance_tests=${PERFORMANCE_TESTS:-false}
+kernels=${KERNELS:-""}
+variants=${VARIANTS:-""}
```
adrienbernede: Variant is overloaded, I checked whether it was a spack variant, but it's a kernel variant, or a perf_variant.
Same in the CI config: VARIANT coexists with SPEC, which is a bit ambiguous. I’m tempted to suggest PERF_ prefix. Would it be to heavy ?

scripts/gitlab/caliper_to_github_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+    if len(sys.argv) != 3:
+        sys.exit("Usage: caliper_to_github_benchmark.py \"input_file_glob_pattern\" output.json")
+
+    pattern = sys.argv[1]
+    output_file = sys.argv[2]
+
+    cali_files = glob.glob(pattern)
+
+    if not cali_files:
+        sys.exit(f"Error: No .cali files found matching pattern: {pattern}")
```
adrienbernede, line 69: Print usage here too?

adrienbernede (approved):

scripts/gitlab/caliper_to_github_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+    reader = cr.CaliperReader()
+    reader.read(input_file)
+
+    benchmarks = []
+
+    for record in reader.records:
+        path = record['path'] if 'path' in record else 'UNKNOWN'
+        total_time_s = float(record[METRIC]) if METRIC in record else 0.0
+        total_time_ms = total_time_s * 1000
+        reps = int(record[REPS_METRIC]) if REPS_METRIC in record else 1
```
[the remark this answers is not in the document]
tdrwenski: I think in the case that the caliper data is non-numeric I would rather let the ValueError/TypeError be thrown than use default values so we can investigate what went wrong.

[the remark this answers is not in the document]
tdrwenski: same comment as above about error handling

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,76 @@
...
+        id: benchmark
+        continue-on-error: true
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: ${{ github.event.inputs.benchmark_name }}
+          tool: 'customSmallerIsBetter'
+          output-file-path: benchmark.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true # TODO remove later, for testing
+          # auto-push: ${{ github.ref == 'refs/heads/develop' }}
```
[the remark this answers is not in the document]
tdrwenski: yes true, but maybe we want to test this workflow works on @adrienbernede's branch and then remove this before merging. I can make a note to remove this later.

.github/workflows/benchmark.yml
tdrwenski: If you want to test that the workflow works and can create the dashboard on your branch, then I say we leave it for a bit longer and then remove it before merging to develop. But good to not forget this.

scripts/gitlab/build_and_test.sh
```diff
@@ -288,21 +293,30 @@ then
...
 then
-    timed_message "Exporting performance data"
-    if [[ -d "${build_dir}/bin" ]]
+
+    cd ${project_dir}
+    mkdir -p ${perf_artifact_dir}
+
+    timed_message "Performance tests for RAJA Perf Suite"
+
+    raja_perf_command="${build_dir}/bin/raja-perf.exe --kernels ${kernels} --variants ${variants} --outdir ${perf_artifact_dir} ${perf_run_opts}"
```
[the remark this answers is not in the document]
tdrwenski: I think we are okay since only developers have rights to the repo can run anything on GitLab and I think there are already other unquoted variables in this script (and ${perf_run_opts} needs to be expanded)

.github/workflows/benchmark.yml
tdrwenski: Good point! The `checks: write` allows the adding of a PR check status. I will remove the `pull-requests: write`-- that is only needed for writing comments to commits/PRs, which I just turned off. `contents: write` is needed to push to the gh-pages branch but also is the one that gives the most power to change things. We only want to do that on `develop` but I don't think you can make permissions conditional. So now I split this into two jobs, one for PRs and one for develop-- there will be a bit of duplication but then each can have the correctly scoped permissions. Let me know what you think!

scripts/gitlab/caliper_to_github_benchmark.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        sys.exit("Usage: caliper_to_github_benchmark.py \"input_file_glob_pattern\" output.json")
+
+    pattern = sys.argv[1]
+    output_file = sys.argv[2]
+
+    cali_files = glob.glob(pattern)
```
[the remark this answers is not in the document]
tdrwenski: Fixed this

.github/workflows/benchmark.yml
adrienbernede: I think this sounds like a great handling of permissions. Merging to develop only happens after review, more permission there is less concerning, so I think it’s a great idea to split despite the duplication.

PR #554, opened 2025-09-23
Feature/grandy1/553 add geometric intersection tests

jmgrandy:
Adds two new tests, INTSC_HEXHEX and INTSC_HEXRECT, to RAJAPerf.

INTSC_HEXHEX computes the volume and moments of intersection between
two hexahedra, each hexahedron represented by 24 triangular facets with a
face centered triangulation.  Each hexahedron is subdivided into eight
12-sided hexahedral subzones, and we compute separately intersections
between pairs of subzones.

INTSC_HEXRECT computes the volume of intersection between a 24 sided
hexahedron and a regular Cartesian rectangular solid.

### DIFF
--- a/src/apps/INTSC_HEXHEX-Cuda.cpp
+++ b/src/apps/INTSC_HEXHEX-Cuda.cpp
@@ -87,7 +87,7 @@ void INTSC_HEXHEX::runCudaVariantImpl(VariantID vid)
   if ( vid == Base_CUDA ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
       constexpr Size_type shmem = 0;
@@ -110,7 +110,7 @@ void INTSC_HEXHEX::runCudaVariantImpl(VariantID vid)
   } else if ( vid == Lambda_CUDA ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
 
@@ -154,7 +154,7 @@ void INTSC_HEXHEX::runCudaVariantImpl(VariantID vid)
   } else if ( vid == RAJA_CUDA ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       RAJA::forall< RAJA::cuda_exec<block_size, true /*async*/> >( res,
         RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
--- a/src/apps/INTSC_HEXHEX-Hip.cpp
+++ b/src/apps/INTSC_HEXHEX-Hip.cpp
@@ -107,7 +107,7 @@ void INTSC_HEXHEX::runHipVariantImpl(VariantID vid)
   if ( vid == Base_HIP ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
       constexpr Size_type shmem = 0;
@@ -130,7 +130,7 @@ void INTSC_HEXHEX::runHipVariantImpl(VariantID vid)
   } else if ( vid == Lambda_HIP ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
 
@@ -174,7 +174,7 @@ void INTSC_HEXHEX::runHipVariantImpl(VariantID vid)
   } else if ( vid == RAJA_HIP ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       RAJA::forall< RAJA::hip_exec<block_size, true /*async*/> >( res,
         RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
--- a/src/apps/INTSC_HEXHEX-Seq.cpp
+++ b/src/apps/INTSC_HEXHEX-Seq.cpp
@@ -49,7 +49,7 @@ void INTSC_HEXHEX::runSeqVariant(VariantID vid,
     case Base_Seq : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         for (Index_type i = ibegin ; i < iend ; ++i ) {
           INTSC_HEXHEX_SEQ ( i, iend ) ;
@@ -68,7 +68,7 @@ void INTSC_HEXHEX::runSeqVariant(VariantID vid,
     case Lambda_Seq : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         for (Index_type i = ibegin ; i < iend; ++i ) {
           intsc_hexhex_lam( i );
@@ -88,7 +88,7 @@ void INTSC_HEXHEX::runSeqVariant(VariantID vid,
       auto res{getHostResource()};
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         RAJA::forall<RAJA::seq_exec>( res,
           RAJA::RangeSegment(ibegin, iend), intsc_hexhex_lam);
--- a/src/apps/INTSC_HEXRECT-Cuda.cpp
+++ b/src/apps/INTSC_HEXRECT-Cuda.cpp
@@ -66,7 +66,7 @@ void INTSC_HEXRECT::runCudaVariantImpl(VariantID vid)
   if ( vid == Base_CUDA ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
       constexpr Size_type shmem = 0;
@@ -84,7 +84,7 @@ void INTSC_HEXRECT::runCudaVariantImpl(VariantID vid)
   } else if ( vid == Lambda_CUDA ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
 
@@ -114,7 +114,7 @@ void INTSC_HEXRECT::runCudaVariantImpl(VariantID vid)
   } else if ( vid == RAJA_CUDA ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       RAJA::forall< RAJA::cuda_exec<block_size, true /*async*/> >( res,
         RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
--- a/src/apps/INTSC_HEXRECT-Seq.cpp
+++ b/src/apps/INTSC_HEXRECT-Seq.cpp
@@ -42,7 +42,7 @@ void INTSC_HEXRECT::runSeqVariant(VariantID vid,
     case Base_Seq : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         for (Index_type i = ibegin ; i < iend ; ++i ) {
           INTSC_HEXRECT_SEQ ( i ) ;
@@ -58,7 +58,7 @@ void INTSC_HEXRECT::runSeqVariant(VariantID vid,
     case Lambda_Seq : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         for (Index_type i = ibegin ; i < iend; ++i ) {
           intsc_hexrect_lam( i );
@@ -75,7 +75,7 @@ void INTSC_HEXRECT::runSeqVariant(VariantID vid,
       auto res{getHostResource()};
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         RAJA::forall<RAJA::seq_exec>( res,
           RAJA::RangeSegment(ibegin, iend), intsc_hexrect_lam);
--- a/src/apps/INTSC_HEXHEX-OMP.cpp
+++ b/src/apps/INTSC_HEXHEX-OMP.cpp
@@ -66,7 +66,7 @@ void INTSC_HEXHEX::runOpenMPVariant(VariantID vid,
     case Base_OpenMP : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         #pragma omp parallel for
         for (Index_type i = ibegin ; i < iend ; ++i ) {
@@ -86,7 +86,7 @@ void INTSC_HEXHEX::runOpenMPVariant(VariantID vid,
     case Lambda_OpenMP : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         #pragma omp parallel for
         for (Index_type i = ibegin ; i < iend ; ++i ) {
@@ -108,7 +108,7 @@ void INTSC_HEXHEX::runOpenMPVariant(VariantID vid,
       auto res{getHostResource()};
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         RAJA::forall<RAJA::omp_parallel_for_exec>( res,
           RAJA::RangeSegment(ibegin, iend), intsc_hexhex_lam);
--- a/src/apps/INTSC_HEXRECT-Hip.cpp
+++ b/src/apps/INTSC_HEXRECT-Hip.cpp
@@ -83,7 +83,7 @@ void INTSC_HEXRECT::runHipVariantImpl(VariantID vid)
   if ( vid == Base_HIP ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
       constexpr Size_type shmem = 0;
@@ -100,7 +100,7 @@ void INTSC_HEXRECT::runHipVariantImpl(VariantID vid)
   } else if ( vid == Lambda_HIP ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
 
@@ -131,7 +131,7 @@ void INTSC_HEXRECT::runHipVariantImpl(VariantID vid)
   } else if ( vid == RAJA_HIP ) {
 
     startTimer();
-    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+    for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
       RAJA::forall< RAJA::hip_exec<block_size, true /*async*/> >( res,
         RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
--- a/src/apps/INTSC_HEXRECT-OMP.cpp
+++ b/src/apps/INTSC_HEXRECT-OMP.cpp
@@ -52,7 +52,7 @@ void INTSC_HEXRECT::runOpenMPVariant(VariantID vid,
     case Base_OpenMP : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         #pragma omp parallel for
         for (Index_type i = ibegin ; i < iend ; ++i ) {
@@ -68,7 +68,7 @@ void INTSC_HEXRECT::runOpenMPVariant(VariantID vid,
     case Lambda_OpenMP : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         #pragma omp parallel for
         for (Index_type i = ibegin ; i < iend ; ++i ) {
@@ -86,7 +86,7 @@ void INTSC_HEXRECT::runOpenMPVariant(VariantID vid,
       auto res{getHostResource()};
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
 
         RAJA::forall<RAJA::omp_parallel_for_exec>( res,
           RAJA::RangeSegment(ibegin, iend), intsc_hexrect_lam);


src/apps/INTSC_HEXRECT-Cuda.cpp
```diff
@@ -0,0 +1,153 @@
...
+        RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
+          {
+            long blksize = block_size ;
+            long blk     = blockIdx.x ;
+            long irec    = blk*blksize + threadIdx.x ;
+
+            int const max_polygon_pts = 10 ;
+            __shared__ double xd_work[ (3 * max_polygon_pts+1) * 64 ] ;
+
+            double *my_qx = xd_work + (3 * max_polygon_pts+1) * threadIdx.x ;
```
artv3: I think we want to use the RAJA constructs here, i.e. RAJA::launch or RAJA::kernel.

src/apps/INTSC_HEXRECT-Sycl.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+  } else if ( vid == RAJA_SYCL ) {
+
+    startTimer();
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      RAJA::forall< RAJA::sycl_exec<work_group_size, true /*async*/> >( res,
+        RAJA::RangeSegment(ibegin, iend), [=] (Index_type i) {
+
+          Index_type i = item.get_global_id(0) + ibegin;
```
artv3: The lambda variable i is being overwritten here.

rhornung67: It's actually a local shadow variable hiding the lambda argument. We need to beef up our CI tests to catch things like this.

jmgrandy: I removed this line of code as it looks redundant.  I have not been able to test with Sycl so this file is currently a placeholder.

rhornung67: @jmgrandy if you click on the corona check in your PR, you will see the compilation failures for the SYCL variant.

src/apps/INTSC_HEXRECT-Cuda.cpp
```diff
@@ -0,0 +1,153 @@
...
+      Int_ptr intsc_t,   // [nrecords] target zones to intersect
+      long const nrecords,  // Number of threads (one thread per record)
+      Real_ptr records )  // output volumes, moments
+{
+  long blksize = block_size ;        // blocksize = 64  must <= nth_per_isc
+  long blk     = blockIdx.x ;
+  long irec    = blk*blksize + threadIdx.x ;   // which thread with offset
+
+  int const max_polygon_pts = 10 ;
+  __shared__ double xd_work[ (3 * max_polygon_pts+1) * 64 ] ;
```
artv3: May be good to define the constant values on top of the files.

src/apps/INTSC_HEXRECT-Seq.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+
+void INTSC_HEXRECT::intscHexRectSeq
+    ( Index_type i, Index_type )
+{
+  INTSC_HEXRECT_DATA_SETUP ;
+
+  long irec = i ;
+  int const max_polygon_pts = 10 ;
+  double xd_work[ 3*max_polygon_pts+1 ] ;
```
artv3: Looks like the size of xd_work is backend dependent. Do you maintain multiple version your app?

src/apps/INTSC_HEXRECT.hpp
```diff
@@ -0,0 +1,200 @@
...
+#define  INTSC_HEXRECT_DATA_SETUP \
+  char *ncord_gpu = m_ncord ; \
+  double const *xdnode = (double const*) m_xdnode ; \
+  double const *ydnode = (double const*) m_ydnode ; \
+  double const *zdnode = (double const*) m_zdnode ; \
+  int const *znlist = (int const*) m_znlist ; \
+  int const *intsc_d = (int const*) m_intsc_d ; \
+  int const *intsc_t = (int const*) m_intsc_t ; \
+  long const nrecords = m_nrecords ; \
+  double *records = (double *)m_records ;
```
MrBurmark: We should use the rajaperf type aliases, Index_type = long, Real_ptr = double*, Int_type = int, etc.
There is no alias for char, so it would be good to add one.
There are no pointer to const aliases but they should be added.

src/apps/INTSC_HEXRECT-Seq.cpp
jmgrandy: xd_work is allocating 31 doubles (248 bytes) of scratch space per thread in all cases.  In Seq it is just an array, and in OMP it is a thread private array so the total usage is 248 bytes * OMP_NUM_THREADS while threads are active.  In Cuda and Hip the workgroup size is 64 and we allocate 64*248 bytes of __shared__ memory, and set a pointer into the array for each thread's scratch space.

In our app this code is deployed on gpu platforms only; we do not execute this code on the CPU in our app.  However running this code on the CPU in RAJAPerf provides a basis of comparison for performance, and additional verification of correctness.

src/apps/INTSC_HEXRECT-Cuda.cpp
jmgrandy: Are you suggesting
static int const max_polygon_pts = 10;
above the function definitions?  I am trying to limit the scope of this variable.

src/apps/INTSC_HEXHEX-Cuda.cpp
```diff
@@ -0,0 +1,186 @@
...
+           Int64_type blksize   = blockDim.x ;
+           Int64_type blk       = blockIdx.x ;
+           Int64_type ith       = blk*blksize + threadIdx.x ;
+           Real_ptr vv_out = (Real_ptr ) vv_int + 8*blk ;
+           INTSC_HEXHEX_BODY; };
+
+      auto intsc_hexhex_fixup_lambda = [=] __device__
+          ( Index_type i )
+         {
+           Int_type ith = blockIdx.x*block_size + threadIdx.x;
```
MrBurmark: Aren't i and ith the same?

src/apps/INTSC_HEXHEX-Cuda.cpp
```diff
@@ -0,0 +1,186 @@
...
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
+
+      auto intsc_hexhex_lambda = [=] __device__
+          ( Index_type i )
+         {
+           __shared__ Real_type vv_reduce[16] ;
+
+           Int64_type blksize   = blockDim.x ;
```
MrBurmark: Can use block_size instead of blockDim.x.

src/apps/INTSC_HEXHEX-Cuda.cpp
```diff
@@ -0,0 +1,186 @@
...
+void INTSC_HEXHEX::runCudaVariantImpl(VariantID vid)
+{
+  const Index_type run_reps = getRunReps();
+  const Index_type ibegin   = 0 ;
+  const Index_type iend     = m_tri_per_intsc * getActualProblemSize() ;
+
+  const Size_type  n_subz_intsc = 8 * getActualProblemSize() ;
+  const Size_type  nisc_stage   = n_subz_intsc ;
+
+  const Size_type  n_szgrp     = ( n_subz_intsc + 7 ) / 8 ;
```
MrBurmark: You can use `RAJA_DIVIDE_CEILING_INT(n_subz_intsc, 8)` here.

src/common/RPTypes.hpp
```diff
@@ -81,6 +81,29 @@ using Size_type = size_t;
...
+using Char_type = char;
+///
+using Char_ptr = Char_type*;
+using Char_const_ptr = Char_type const*;
+
+using Uchar_type = unsigned char;
+///
+using Uchar_ptr = Uchar_type*;
+
+using Int64_type = std::int64_t ;
```
MrBurmark: Can this use Index_type or is this used for something that explicitly needs 64 bits?

src/common/RPTypes.hpp
```diff
@@ -120,6 +143,9 @@ using Real_type = float;
 #endif
 
 using Real_ptr = Real_type*;
+using Real_ptr2 = Real_type**;
```
MrBurmark: What do people think about `Real_ptr_ptr` compared vs `Real_ptr2`.

src/apps/INTSC_HEXRECT-Seq.cpp
MrBurmark: So this is a per thread scratchpad of values. If we think about this as a tiled loop then the size depends on parallelism in the inner tile, 1 for sequential and OMP and block_size for gpus. For GPUs I assume that you're doing this either because you do dynamic indexing of this memory or you ran into register pressure and found a way to save registers.

src/apps/INTSC_HEXRECT-Cuda.cpp
MrBurmark: It looks like this is used as a constant across all variants, it would make sense to me to define it as `inline constexpr int max_polygon_pts = 10 ;` in the hpp file.

src/apps/INTSC_HEXRECT-Cuda.cpp
```diff
@@ -0,0 +1,153 @@
...
+      Int_ptr intsc_t,   // [nrecords] target zones to intersect
+      Int64_type const nrecords,  // Number of threads (one thread per record)
+      Real_ptr records )  // output volumes, moments
+{
+  Int64_type blksize = block_size ;    // blocksize = 64  must <= nth_per_isc
+  Int64_type blk     = blockIdx.x ;
+  Int64_type irec    = blk*blksize + threadIdx.x ;   // which thread with offset
+
+  Int_type const max_polygon_pts = 10 ;
+  __shared__ Real_type xd_work[ (3 * max_polygon_pts+1) * 64 ] ;
```
MrBurmark: This 64 should be block_size?

src/apps/INTSC_HEXHEX-Cuda.cpp
MrBurmark: Generally we try to use the provided index i instead of using anything backend specific like threadIdx in the lambda and RAJA variants.

src/apps/INTSC_HEXRECT-Seq.cpp
jmgrandy: Yes, both.  We do need dynamic indexing and using a local stack array can cause spills or lower occupancy.

src/common/RPTypes.hpp
jmgrandy: The global thread number (such as ith) can range up to 2^32 threads per kernel launch on the
MI300 which is a factor of two above INT_MAX and hence needs to be 64 bit in the app code.

The limit of 2^32 threads is about 7.4 million standard intersections (576 GPU threads per intersection) and
is quite realistic in the app code as this would take about 1 to 1.5 sec on an MI300.

As a matter of practice I kept some other indexing variables such as block number at 64 bit.

MrBurmark: I would still use Index_type here as the variable is intending to hold indices. RAJAPerf does use a 64-bit integer for this by default and in the case that we try using a 32-bit integer in RAJAPerf we understand that some kernels will end up limited by the range of representable values.

pearce8 (commented):
@rhornung67 @MrBurmark Is there a place @jmgrandy's description of the kernels in the PR description should live beyond the lifetime of this PR?  It would be useful to keep these so we can find them later.

src/apps/INTSC_HEXRECT-Cuda.cpp
jmgrandy: see commit "Replaced multiple definitions of max_polygon_pts..."

src/apps/INTSC_HEXRECT-Cuda.cpp
jmgrandy: See "Use index variable instead of blockIdx..." commit.

src/apps/INTSC_HEXHEX-Cuda.cpp
jmgrandy: See commit "Use index variable instead of blockIdx...".

src/apps/INTSC_HEXHEX-Cuda.cpp
jmgrandy: See commit "Use index variable instead of blockIdx...".

src/apps/INTSC_HEXHEX-Cuda.cpp
jmgrandy: See commit "Convert redundant ceiling divisions..."

src/common/RPTypes.hpp
jmgrandy: See commit "Replaced Int64_type with Index_type..."

src/apps/INTSC_HEXRECT-Cuda.cpp
jmgrandy: See commit "Replace number 64..."

scripts/lc-builds/toss4_icpx.sh
```diff
@@ -40,8 +40,8 @@ source /usr/tce/packages/intel/intel-${COMP_VER}/setvars.sh
 
 cmake \
   -DCMAKE_BUILD_TYPE=Release \
-  -DCMAKE_CXX_COMPILER=icpx \
-  -DCMAKE_C_COMPILER=icx \
+  -DCMAKE_CXX_COMPILER=/usr/tce/packages/intel/intel-${COMP_VER}/bin/icpx \
+  -DCMAKE_C_COMPILER=/usr/tce/packages/intel/intel-${COMP_VER}/bin/icx \
```
MrBurmark: This diff shouldn't exist.

src/apps/INTSC_HEXRECT.hpp
```diff
@@ -0,0 +1,203 @@
...
+#define  INTSC_HEXRECT_DATA_SETUP \
+  Char_ptr ncord_gpu = m_ncord ; \
+  Real_const_ptr xdnode = (Real_const_ptr) m_xdnode ; \
+  Real_const_ptr ydnode = (Real_const_ptr) m_ydnode ; \
+  Real_const_ptr zdnode = (Real_const_ptr) m_zdnode ; \
+  Int_const_ptr znlist  = (Int_const_ptr) m_znlist  ; \
+  Int_const_ptr intsc_d = (Int_const_ptr) m_intsc_d ; \
+  Int_const_ptr intsc_t = (Int_const_ptr) m_intsc_t ; \
+  Index_type const nrecords = m_nrecords ; \
+  Real_ptr records = (Real_ptr )m_records ;
```
MrBurmark: We probably don't need to explicitly cast these to const pointers.

src/apps/INTSC_HEXHEX.hpp
```diff
@@ -0,0 +1,154 @@
...
+  void intscHexHexOMP_Target ( Index_type i, Index_type iend ) ;
+
+  void check_intsc_volume_moments
+      ( Index_type const n_intsc, Real_const_ptr vv ) ;
+
+  static constexpr size_t m_tri_per_intsc = 576 ;
+  static const size_t default_gpu_block_size = 64;
+  using gpu_block_sizes_type = integer::make_gpu_block_size_list_type<default_gpu_block_size>;
+
+  VariantID m_vid ;
```
MrBurmark: Why is this here?

scripts/lc-builds/toss4_icpx.sh
jmgrandy: It would not compile without this change, perhaps I am supposed to load a module.
However toss4_clang.sh has the full path in specifying the compiler so this proposed
change does the same thing for icpx.  Please clarify what is the correct process for
compiling with icpx.

src/apps/INTSC_HEXRECT.hpp
jmgrandy: I am casting these pointers to const pointers.

scripts/lc-builds/toss4_icpx.sh
MrBurmark: The file that is sourced above should be setting up the environment. The full path is not consistent between all compiler versions so I removed it. Which compiler version are you using that has this problem?

src/apps/INTSC_HEXRECT.hpp
MrBurmark: That should not be necessary, its just things like `int* -> const int*` right?

src/apps/INTSC_HEXHEX.hpp
jmgrandy: This kernel and INTSC_HEXRECT are specifically designed and constrained to run with a block size of 64 and will not work with larger block sizes.

MrBurmark: I meant `m_vid`, I guess hithub is trying to be helpful by showing more context.

jmgrandy: I made m_vid a member so we could access it in the error checking routine and print out which variant gave an incorrect result.

src/apps/INTSC_HEXRECT.hpp
jmgrandy: It is not strictly necessary for execution but the const cast helps detect inadvertent modifications when compiling.  Shall I remove these casts to const pointers?

scripts/lc-builds/toss4_icpx.sh
jmgrandy: With the toss4_icpx.sh script int develop,

  ./scripts/lc-builds/toss4_icpx.sh 2025.2.0-magic

./scripts/lc-builds/toss4_icpx.sh: line 39: /usr/tce/packages/intel/intel-2025.2.0-magic/setvars.sh: No such file or directory

    icpx

  is not a full path and was not found in the PATH.

MrBurmark: Ok, I'll see if I can fix the path to the script in another PR.

src/apps/INTSC_HEXRECT.hpp
MrBurmark: Ya, please remove these.

src/apps/INTSC_HEXHEX.hpp
MrBurmark: Can you change that to pass the variant in as an argument?

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  Real_ptr dcoord ;   // donor  coordinates [24]
+  Real_ptr tcoord ;   // target coordinates [24]
+
+  // Expanded donor and target coordinates on host
+  Real_ptr ds_h, ts_h ;
+
+  do {
+    using namespace detail ;
```
MrBurmark: Please use allocDataForInit here to setup data on the host.

MrBurmark: It returns an object that will copy the host data to the right place at the end of scope.

MrBurmark: Generally there should not be any direct usage of methods in the detail namespace here.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,303 @@
...
+      Real_type dym = vv[ 4*k + 2 ] - vy ;
+      Real_type dzm = vv[ 4*k + 3 ] - vz ;
+
+      // Print an error message if a volume or moment is incorrect.
+      if ( ( dv*dv   > tolsqv ) ||
+           ( dxm*dxm > tolsqx ) ||
+           ( dym*dym > tolsqy ) ||
+           ( dzm*dzm > tolsqz ) ) {
+
+        printf ( "%s %s %s.\n", tst,
```
MrBurmark: Please use `getCout() <<` here. That only prints from rank 0 so you can let all the ranks check and then only rank 0 will actually print what it gets to the screen.

scripts/lc-builds/toss4_icpx.sh
MrBurmark: The non-magic compilers work.

src/apps/INTSC_HEXHEX.cpp
jmgrandy: See commit "Replace printf"

src/apps/INTSC_HEXHEX.cpp
jmgrandy: See commit "Replace detail::allocHostData".

src/apps/INTSC_HEXHEX.hpp
jmgrandy: See commit "Remove m_vid member".

src/apps/INTSC_HEXRECT.hpp
jmgrandy: See commit "Remove some casts".

scripts/lc-builds/toss4_icpx.sh
jmgrandy: See commit "Restore toss4_icpx.sh"

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,309 @@
...
+    ztzone[i] = zdzone[i] + m_shift ;
+  }
+
+  Real_ptr dcoord ;   // donor  coordinates [24]
+  Real_ptr tcoord ;   // target coordinates [24]
+
+  // Expanded donor and target coordinates on host
+  Real_ptr ds_h, ts_h ;
+
+  do {
```
MrBurmark: Why use `do {} while (0);` here? A scope `{}` accomplishes the same thing.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  allocAndCopyHostData ( m_dsubz, ds_h, 24L*n_subz_intsc, vid ) ;
+  allocAndCopyHostData ( m_tsubz, ts_h, 24L*n_subz_intsc, vid ) ;
+
+  do {
+    deallocData ( ds_h, Base_Seq ) ;
+    deallocData ( ts_h, Base_Seq ) ;
+    deallocData ( dcoord, Base_Seq ) ;
+    deallocData ( tcoord, Base_Seq ) ;
```
MrBurmark: How about using `allocDataForInit` on `m_dsubz` and `m_tsubz` and filling them without using any intermediate allocations?

MrBurmark: See an example of `allocDataForInit` here [`algorithm/HISTOGRAM.cpp:73`](https://github.com/LLNL/RAJAPerf/blob/5fafb94b0f31aa14a9060ad6b514b38edc7caca4/src/algorithm/HISTOGRAM.cpp#L73).

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    allocDataForInit ( ds_h, 24L*n_subz_intsc, Base_Seq ) ;
+    allocDataForInit ( ts_h, 24L*n_subz_intsc, Base_Seq ) ;
+
+  } while ( false ) ;
+
+  //  Repeat the same calculation n_subz_intsc times, expand the
+  //  same donor and target zones.
+  for ( Index_type k=0 ; k < n_subz_intsc ; ++k ) {
+    memcpy ( ds_h + 24L*k, dcoord, 24*sizeof(Real_type) ) ;
```
MrBurmark: Please use copyData instead of memcpy https://github.com/LLNL/RAJAPerf/blob/5fafb94b0f31aa14a9060ad6b514b38edc7caca4/src/common/KernelBase.hpp#L353.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,309 @@
...
+      // Print an error message if a volume or moment is incorrect.
+      if ( ( dv*dv   > tolsqv ) ||
+           ( dxm*dxm > tolsqx ) ||
+           ( dym*dym > tolsqy ) ||
+           ( dzm*dzm > tolsqz ) ) {
+
+        Size_type paragraph_max = 1000 ;
+
+        std::string sbuf  ;
+        Char_ptr buf = new Char_type [ paragraph_max ] ;
```
MrBurmark: It would be easier to use a std container like std::vector or std::array here.
```cpp
std::array<Char_type, paragraph_max> buf{{'\0'}};
std::vector<Char_type> buf(paragraph_max, '\0');
```

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,309 @@
...
+  // One standard intersection is 8 subzone intersections.
+  Index_type n_std_intsc  = getActualProblemSize() ;
+  Index_type n_subz_intsc = 8L * n_std_intsc ;
+
+  copyData ( DataSpace::Host, m_vv,
+             getDataSpace(vid), m_vv_out, 4L*n_subz_intsc ) ;
+
+  check_intsc_volume_moments ( n_subz_intsc, m_vv, vid ) ;
+
+  deallocData ( m_vv, Base_Seq ) ;
```
MrBurmark: Anything allocated in setup should be deallocated in teardown.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,309 @@
...
+      Real_const_ptr vv,   // computed volumes, moments on the host
+      VariantID vid )   // Print variant name in case of error
+{
+  Int_type rank = 0;
+#if defined(RAJA_PERFSUITE_ENABLE_MPI)
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+#endif // RAJA_PERFSUITE_ENABLE_MPI
+
+  // Check on rank 0, other ranks are identical.
+  if ( rank == 0 ) {
```
MrBurmark: I would prefer to simplify this code a bit and let all ranks do this, then only rank 0 will actually print due to the implementation of getCout.

src/apps/INTSC_HEXHEX-Seq.cpp
```diff
@@ -0,0 +1,130 @@
...
+  if ( i == 0 ) {   // initialize the accumulation
+    Index_type n_std_intsc = m_nthreads / m_tri_per_intsc ;
+    Index_type vv_len = 32L * n_std_intsc ;
+    for ( Index_type k = 0 ; k < vv_len ; ++k ) {
+      m_vv_out[k] = 0.0 ;
+    }
+  }
+
+  Real_ptr tsubz = m_tsubz ;
+  Real_ptr dsubz = m_dsubz ;
```
MrBurmark: We normally pull out the data members out and initialize things in a SETUP macro.

src/apps/INTSC_HEXHEX.cpp
jmgrandy: See "Replaced do .. while" commit.
https://github.com/LLNL/RAJAPerf/pull/554/commits/3acd94ec60a1922cd276adae1049e7e3ea54e7de

src/apps/INTSC_HEXHEX.cpp
jmgrandy: See "Replace memset and memcpy" commit.
https://github.com/LLNL/RAJAPerf/pull/554/commits/0bce8eeaed95c98b8bcacb49152c1238521f785b

src/apps/INTSC_HEXHEX.cpp
jmgrandy: See "Error reporting vector" and "Removed rank 0" commits.
https://github.com/LLNL/RAJAPerf/pull/554/commits/67a34eada29720f6006a6d6edde236e232ca3cbc
https://github.com/LLNL/RAJAPerf/pull/554/commits/491e900ea45bdf51e082181fd532f68fa4f2e804

src/apps/INTSC_HEXHEX.cpp
jmgrandy: See "Moved deallocations" commit.
https://github.com/LLNL/RAJAPerf/pull/554/commits/436bc9de0243ed28f541d254c798bf21b1c8484f

src/apps/INTSC_HEXHEX.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/491e900ea45bdf51e082181fd532f68fa4f2e804

src/apps/INTSC_HEXHEX-Seq.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/472041773621cd78ac9b6c290f2179c9da9b20fc

src/apps/INTSC_HEXHEX.cpp
jmgrandy: I have a question about this.  In HISTOGRAM.cpp m_bins is being allocated with the VariantID
so you might be allocating on the device but you are writing to it (m_bins[i] = bin;) in host code.
What happens when there are separate memory spaces?  Does allocDataForInit always allocate on the
host; if so why are we passing in VariantID?  Is it managed memory?

MrBurmark: Sorry there is no documentation for this. The *ForInit versions allocate in a memory space that is host accessible and then copy that to the final space when their return value is destroyed. So the idea is that with minimal code, one function call and a local variable, you can get something that can be setup in host code then moved to the right memory space automatically.

src/apps/INTSC_HEXHEX-Hip.cpp
```diff
@@ -0,0 +1,211 @@
...
+#define WARPSIZE 64
+#define __shfl_xor_sync(mask,val,n) __shfl_xor(val,n)
+
+namespace rajaperf
+{
+namespace apps
+{
+
+template < Size_type block_size >
+__launch_bounds__(block_size,3)
```
MrBurmark, line 30: `,3`?

src/apps/INTSC_HEXHEX-Hip.cpp
```diff
@@ -0,0 +1,211 @@
...
+    RPlaunchHipKernel( (intsc_hexhex_hip<block_size>),
+                       grid_size1, block_size,
+                       shmem, res.get_stream(),
+                       m_dsubz, m_tsubz,
+                       0UL, m_vv_int ) ;
+    RPlaunchHipKernel( (intsc_hexhex_hip_fixup_vv_64to72<block_size>),
+                       grid_size1, block_size,
+                       shmem, res.get_stream(),
+                       m_vv_int, 0UL, m_vv_out ) ;
+  }
```
MrBurmark, line 104: The warmup kernel problem is a broader problem that needs to be addressed. I have an idea of a way to handle this in general, so I'm not sure that it should be addressed in this PR. I would leave this out for now.

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: Yes, I found during the January hackathon that we got better performance with lower occupancy than the default 4, and three was the best, with Rocm 6.2.  With the more recent Rocm 6.4 the compiler the kernel is much slower due to register usage so the launch bounds probably don't have much effect.

src/apps/INTSC_HEXRECT.hpp
```diff
@@ -0,0 +1,205 @@
...
+  Real_ptr m_records_h ;       // volumes, moments on the host
+
+  Real_type m_xd0, m_yd0, m_zd0 ;    // donor corner
+  Real_type m_x0, m_y0, m_z0 ;       // corner of target mesh
+  Real_type m_sep ;                  // target mesh pitch (separation)
+  Real_type m_sep1x, m_sep1y, m_sep1z ;   // donor mesh zone widths
+  Int_type  m_x_scl_offs, m_y_scl_offs, m_z_scl_offs ;  // donor scaled offsets
+  Int_type  m_ndx, m_ndy, m_ndz ;     // donor mesh dimensions (are equal)
+
+  FILE *m_f_hexrect ;  // output file
```
MrBurmark: Why is there an output file?

src/apps/INTSC_HEXHEX-Seq.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+  // Volumes directly to vv_out on the CPU.
+  Real_ptr vv_out = m_vv_out + 4L*ipair;
+
+  //   Save results for this triangle, for the subzone pair intersection.
+  vv_out[0] += vv_hi + vv_lo ;
+  vv_out[1] += vx_hi + vx_lo ;
+  vv_out[2] += vy_hi + vy_lo ;
+  vv_out[3] += vz_hi + vz_lo ;
+}
```
MrBurmark: Can you put this in a macro instead of a function? That should let you pull the SETUP part out of this code.
The idea of the Lambda variants is to see that compilers can optimize out the lambda and its function call compared to the base implementation where there is no function call.

src/apps/intsc_hexrect_body.hpp
```diff
@@ -0,0 +1,622 @@
...
+      Int_type &abovez,
+      Int_type &clip )          // clip - 12 out of 24 triangles is common
+{
+  inside = abovez = clip = 0 ;   // initialize masks.
+
+  //  target z spacing
+  Real_type const dzt = ( zt1 > zt0 ) ? ( zt1 - zt0 ) : 0.0 ;
+
+  Real_type a11, a22 ;    // Transformation to unit target frame 0<(x,y)<1
+  do {
```
MrBurmark: Are these using `do {} while (false)` because they were in macros before? Can you change all of the `do {} while (false)` into just scopes `{}`?

src/apps/intsc_hexhex_body.hpp
```diff
@@ -0,0 +1,492 @@
...
+        vv_thr, vx_thr, vy_thr, vz_thr ) ;
+}
+
+}  // closing brace for rajaperf namespace
+
+
+#define INTSC_HEXHEX_BODY_SEQ \
+  Index_type const n_dsz_tris = 12 ; \
+  Index_type const n_tsz_tets = 6 ; \
+  Index_type const nth_per_isc = n_dsz_tris * n_tsz_tets ; \
```
MrBurmark: Could these be constexpr static class member variables or defines?

src/apps/intsc_hexhex_body.hpp
```diff
@@ -0,0 +1,492 @@
...
+  Index_type ipair   = ith / nth_per_isc ; \
+  Int_type dfacet  = ( ith / n_tsz_tets ) % n_dsz_tris ; \
+  Int_type ttet    = ith % n_tsz_tets ; \
+  Index_type pair_base_thr = ipair * nth_per_isc ; \
+  Index_type blk_base = blk * blksize ; \
```
MrBurmark: I was unclear before when I said that everything should be structured like the GPU implementations. I thought that there would be sequential optimizations like caching or something like that. I would implement this as multiple loops to avoid the divides and mods here. In the GPU loop there may have to be different calculations but we generally try to express things as nested loops so we can use our abstractions.

src/apps/INTSC_HEXHEX.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/666c0c4958c692df54cd508159f34787b2ec5cbf
https://github.com/LLNL/RAJAPerf/pull/554/commits/c034c64aab0662505c97a91a749abf00c417ce81
Using allocDataForInit reduced code volume as advertised.

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/c034c64aab0662505c97a91a749abf00c417ce81
The first kernel execution without warmup is about 3x slower for INTSC_HEXHEX and about 8x slower for INTSC_HEXRECT.  These kernels are rather large and for the problem sizes we are using the kernel execution times on the MI300 after warmup are about 2.5 ms for INTSC_HEXHEX and 4.3 ms for INTSC_HEXRECT.

src/apps/INTSC_HEXRECT.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/3749f58f20f827ff2ee81be2450b9fbc74e4a030
We had an output file for debugging during development and stopped using it.

src/apps/INTSC_HEXHEX-Seq.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/385be38d3327332dedc9c6d24033361d36ad0530

src/apps/intsc_hexrect_body.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/385be38d3327332dedc9c6d24033361d36ad0530
Yes, I had overlooked these.

src/apps/intsc_hexhex_body.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/385be38d3327332dedc9c6d24033361d36ad0530
Made these constexpr; they are within the kernel body.

src/apps/INTSC_HEXRECT-Sycl.cpp
jmgrandy: It is compiling successfully on SYCL now.

We still need to get the INTSC_HEXHEX kernel to work there; hopefully that can be a separate PR.

src/apps/INTSC_HEXHEX-Hip.cpp
MrBurmark: Wow, we could make this an additional tuning parameter then if it makes a big difference.

src/apps/INTSC_HEXHEX-Hip.cpp
MrBurmark: Wow, do you need to warmup both? I was hoping that just launching an empty kernel in the translation unit once would be enough to capture the cost of loading the device code. Do you think there is more overhead associated with these kernels?

src/apps/INTSC_HEXHEX-OMPTarget.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Volumes directly to vv_out on the CPU.
+    Real_ptr vv_out = m_vv_out + 4L*ipair;
+
+    //   Save results for this triangle, for the subzone pair intersection.
+    vv_out[0] += vv_hi + vv_lo ;
+    vv_out[1] += vx_hi + vx_lo ;
+    vv_out[2] += vy_hi + vy_lo ;
+    vv_out[3] += vz_hi + vz_lo ;
+  }
+}
```
MrBurmark: Can you move these functions into macros for the other variants?

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/a1b4d19cace283d51b8b02d454d2ea1b392f2199

I have removed the warmup kernels.  I believe I had a copy link error in my message above https://github.com/LLNL/RAJAPerf/pull/554#discussion_r2400105608

src/apps/INTSC_HEXHEX-OMPTarget.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/a16b0d6b0bee25bf3d753768952fed0ac299e4a4

src/apps/intsc_hexhex_body.hpp
MrBurmark: Regarding differences between cpu and gpu variants of the kernel this kernel is already inconsistent as the cpu variants have one loop while the gpu versions use two loops. These kind of differences are what I was hoping to avoid. We count things like the number of kernel launches/parallel loop nests so that shouldn't differ between variants.

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/75f7a176bda7f90a5074af9f7fa1b80137f3a632

After consideration I would like to keep the warmup kernels, due to the large effect of initializing the kernel on the execution time of the first launch only.  I hope we can leave it in for now until the more formal way is available in a future PR.

src/apps/intsc_hexhex_body.hpp
jmgrandy: Ok I believe you are asking about the fixup kernel in INTSC_HEXHEX; I will add a loop for this into the cpu variants.

src/common/RPTypes.hpp
artv3: I think ``Real_ptr_ptr`` is more clear

src/apps/intsc_hexhex_body.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/eab1648cbf3b74cd99c89a1614edda748321045e

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/75f7a176bda7f90a5074af9f7fa1b80137f3a632
I restored the warmup kernels due to the large kernel initialization time in the first call to the kernel.
For INTSC_HEXHEX the warmup sends nisc_stage=0 and thus returns immediately doing no work.

src/common/RPTypes.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/84f9d57bf53abeca7288a54635c41c8df929b962

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,283 @@
...
+  setVariantDefined( Base_CUDA );
+  setVariantDefined( Lambda_CUDA );
+  setVariantDefined( RAJA_CUDA );
+
+  setVariantDefined( Base_HIP );
+  setVariantDefined( Lambda_HIP );
+  setVariantDefined( RAJA_HIP );
+
+  setVariantDefined( Base_SYCL );
+  setVariantDefined( RAJA_SYCL );
```
MrBurmark: Don't say its defined if it isn't defined.

src/apps/INTSC_HEXHEX.hpp
```diff
@@ -0,0 +1,202 @@
...
+    (gpu_block_size + min_warp_size - 1) / min_warp_size ;
+
+// Maximum number of pairs represented in a gpu block.
+// With gpu_block_size < tri_per_pair at most two pairs represented in a block.
+//
+  static Index_type constexpr max_pairs_per_block = 2 ;
+
+  // Number of intermediate computed values per gpu block
+  static Index_type constexpr n_vvint_per_block =
+      max_pairs_per_block * nvals_per_pair ;
```
MrBurmark: spacing

src/apps/INTSC_HEXHEX.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+// 72 data entries in group in intermediate results (before fixup).
+static Index_type n_vvint_per_group = n_vvint_per_block * blks_per_group ;
+
+  // shared memory size for reduction within a block.
+  // The two distinct subzone pairs in a block are distinct.
+  // For each subzone pair, have reduction values per warp.
+  // len_vv_reduce = 16.
+  static Index_type constexpr len_vv_reduce =
+      max_warps_per_block * max_pairs_per_block * nvals_per_pair ;
```
MrBurmark: spacing

src/apps/INTSC_HEXHEX.hpp
```diff
@@ -0,0 +1,202 @@
...
+}
```
MrBurmark:
```suggestion
} // end namespace rajaperf
```

src/apps/INTSC_HEXHEX_BODY.hpp
```diff
@@ -0,0 +1,547 @@
...
+    j = jp ;
+  }
+
+
+  if ( j1 >= 0 ) {     // Set linked list for crossover points.
+    next[j1] = jr1 ;
+    next[jr1] = jr2 ;
+    next[jr2] = ( ( clast < 0 ) || ( c00 < 0 ) ) ? -1 : jj2 ;
+  }
+  return ;
```
MrBurmark: Why have an empty return at the end of the function?

src/apps/INTSC_HEXHEX_BODY.hpp
```diff
@@ -0,0 +1,547 @@
...
+
+    j2 = next[j2] ;
+  }
+}
+
+
+
+RAJA_HOST_DEVICE
+RAJA_INLINE void cuda_intsc_tri_tet
+    ( Real_type const xdt[3],    // donor triangle coordinates
```
MrBurmark: If you pass by reference it will check that the size matches, which you may or may not want.
```suggestion
    ( Real_type const (&xdt)[3],    // donor triangle coordinates
```

src/apps/INTSC_HEXHEX_BODY.hpp
```diff
@@ -0,0 +1,547 @@
...
+  Index_type ith = i ; \
+  Index_type blk = ith / blksize ; \
+  Real_ptr vv_int_p = vv_int + n_vvint_per_block * blk ; \
+  if ( i == 0 ) { \
+    Index_type gsize = iend / blksize ; \
+    Index_type vv_int_len = n_vvint_per_block * gsize ;  \
+    for ( Index_type k = 0 ; k < vv_int_len ; ++k ) { \
+      vv_int_p[k] = 0.0 ; \
+    } \
+  } \
```
MrBurmark, line 478: why is this different from the parallel implementations?

src/apps/INTSC_HEXRECT.cpp
```diff
@@ -0,0 +1,667 @@
...
+      Real_type const sep )    // plane pitch (separation)
+{
+  Int_type nx = ndx + 2 ;     // number of target planes
+  Int_type ny = ndy + 2 ;
+  Int_type nz = ndz + 2 ;
+  Int_type ntx = ndx + 1 ;    // number of target zones each direction
+  Int_type nty = ndy + 1 ;
+  Int_type ntz = ndz + 1 ;
+
+  planes[0] = new Real_type[nx+ny+nz] ;
```
MrBurmark: Use allocData.

src/apps/INTSC_HEXHEX_BODY.hpp
jmgrandy: vv_int contains contributions to volume and moments accumulated over blocks of 64 threads.  In Sequential and OpenMP we loop over individual contributions, and must initialize vv_int prior to computing contributions.  On the GPU we perform the butterfly sum and write the values summed over the block directly into vv_int hence no need to initialize vv_int.

src/apps/INTSC_HEXRECT.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/1e6ca97190b49e4aa31811f5098772eaa5232a83

src/apps/INTSC_HEXHEX.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/649e54f3295b16adbe91e759474e3d61d673f7ae

src/apps/INTSC_HEXHEX.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/649e54f3295b16adbe91e759474e3d61d673f7ae

src/apps/INTSC_HEXHEX.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/649e54f3295b16adbe91e759474e3d61d673f7ae

src/apps/INTSC_HEXHEX_BODY.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/649e54f3295b16adbe91e759474e3d61d673f7ae

src/apps/INTSC_HEXHEX.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/772ef02a0004f6c6742a683864ba7698960a67f7

src/apps/INTSC_HEXHEX_BODY.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/8390ddb7582d712f025bf920140f550e7c193065

src/apps/INTSC_HEXHEX_BODY.hpp
MrBurmark: It looks like this initializes vv_int globally here? Can it just do it locally?

src/apps/INTSC_HEXHEX-Cuda.cpp
```diff
@@ -0,0 +1,195 @@
...
+  ( Real_ptr const dsubz,
+    Real_ptr const tsubz,
+    Size_type  const nisc_stage,
+    Real_ptr vv_int )
+{
+  __shared__ Real_type vv_reduce[len_vv_reduce] ;
+
+  Index_type blksize = block_size ;   // blocksize = 64  must <= tri_per_pair
+  Index_type blk     = blockIdx.x ;
+  Index_type ith     = blk*blksize + threadIdx.x ;  // which thread with offset
```
artv3, line 40: One thing I wanted I wanted to ask is if the launch bounds and block_size values are done in this manner in the application code? These optimizations can make the raja perf versions faster than what is used in the application code. If not, perhaps we can have them as "tunings"?

src/apps/INTSC_HEXHEX-Hip.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  } else if ( vid == RAJA_HIP ) {
+
+    startTimer();
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      RAJA::forall< RAJA::hip_exec<block_size, true /*async*/> >( res,
+        RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
+          {
+            __shared__ Real_type vv_reduce[len_vv_reduce] ;
```
artv3:
```suggestion
            RAJA_TEAM_SHARED Real_type vv_reduce[len_vv_reduce] ;
```
Minor suggestion

src/apps/INTSC_HEXHEX-Sycl.cpp
```diff
@@ -0,0 +1,133 @@
...
+      //   Index_type blk     = i / blksize ;
+      //   Index_type thridx  = i % work_group_size ;
+      //   Real_ptr vv_int_p = (Real_ptr) vv_int + 8*blk ;
+      //   INTSC_HEXHEX_BODY;
+      // });
+
+      // RAJA::forall< RAJA::sycl_exec<work_group_size, true /*async*/> >( res,
+      //   RAJA::RangeSegment(ibegin, iend_fixup), [=] (Index_type i) {
+      //   FIXUP_VV_BODY ;
+      //   }
```
artv3: What happened here? why is it commented out?

src/apps/INTSC_HEXRECT-Cuda.cpp
```diff
@@ -0,0 +1,148 @@
...
+    startTimer();
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      RAJA::forall< RAJA::cuda_exec<block_size, true /*async*/> >( res,
+        RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i)
+          {
+            Index_type irec    = i ;
+            Index_type thridx  = i % block_size ;
+
+            __shared__ Real_type xd_work[(3 * max_polygon_pts+1) * block_size];
```
artv3, line 125:
```suggestion
            RAJA_TEAM_SHARED Real_type xd_work[(3 * max_polygon_pts+1) * block_size];
```

src/apps/INTSC_HEXHEX_BODY.hpp
jmgrandy: This segment of code is executed only for the Sequential variant.  For the OpenMP variant each thread initializes the values for the groups it computes (the initialization is thread local).

src/apps/INTSC_HEXHEX-Sycl.cpp
jmgrandy: It was failing the Sycl compile test due to __syncthreads and __shfl_xor_sync and I was unable to test and debug replacements since I do not have an account on a Sycl platform.  I hope to get the Sycl variant working in the future.

src/apps/INTSC_HEXHEX-Cuda.cpp
jmgrandy: Yes, the kernel is written specifically for a block size of 64 due to its distribution of work.  I found that the launch bounds helped performance of this kernel during the January hackathon, so I plan to deploy it in the app code unless future Rocm versions or future machines show different performance results.

src/apps/INTSC_HEXRECT-Cuda.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/7410f29645552087177efa7aef14adf8f9b3f756

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/7410f29645552087177efa7aef14adf8f9b3f756

src/apps/INTSC_HEXHEX-Cuda.cpp
MrBurmark: It sounds like it may be useful to have the option of sweeping across a variety of launch bound options. When we merge this we should make an issue with all the changes we intend to make to these kernels.

src/apps/INTSC_HEXHEX-Sycl.cpp
MrBurmark: I would remove it entirely if its not ready to go. You can make a branch if there's anything you'd like to keep.

jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/f93d0d14653ecc4efe3cbb0869ffa87059773db1

MrBurmark: Please remove the whole file and don't declare any sycl methods. See this file as an example https://github.com/LLNL/RAJAPerf/blob/572956ab398fa8218734d6a7fe58d120be2d7f51/src/apps/NODAL_ACCUMULATION_3D.hpp#L99

jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/825cb8793a0b9511172c72d50cb42b2f6a111db3

src/apps/INTSC_HEXRECT-Sycl.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/825cb8793a0b9511172c72d50cb42b2f6a111db3

src/apps/INTSC_HEXHEX-Cuda.cpp
```diff
@@ -0,0 +1,195 @@
...
+           Index_type ith       = i ;
+           Index_type thridx    = i % block_size ;
+
+           Real_ptr vv_int_p = (Real_ptr ) vv_int + 8*blk ;
+           INTSC_HEXHEX_BODY; };
+
+      auto intsc_hexhex_fixup_lambda = [=] __device__
+          ( Index_type i )
+         {
+           FIXUP_VV_BODY ; } ;
```
rhornung67:
```suggestion
      auto intsc_hexhex_fixup_lambda = [=] __device__ ( Index_type i ) {
          FIXUP_VV_BODY ; 
      } ;
```

I think code formatting and alignment like this is easier to read. It's also consistent with implementations in other files throughout RAJA Perf.

src/apps/INTSC_HEXHEX-Cuda.cpp
```diff
@@ -0,0 +1,195 @@
...
+    stopTimer();
+
+  } else if ( vid == Lambda_CUDA ) {
+
+    startTimer();
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
+
+      auto intsc_hexhex_lambda = [=] __device__
```
rhornung67: Please see comment on code formatting below.

src/apps/INTSC_HEXHEX-Hip.cpp
```diff
@@ -0,0 +1,215 @@
...
+  const Size_type  n_szpairs   = n_subz_intsc ;
+
+  auto res{getHipResource()};
+
+  INTSC_HEXHEX_DATA_SETUP;
+
+  //  Insert a warmup call to the kernel in order to remove the
+  //  time of initialization that affects the first call to the kernel.
+  //   The warmup calls send n_subz_intsc=0 and hence return immediately.
+  Bool_type const do_warmup = true ;
```
rhornung67, line 90: @jmgrandy @MrBurmark I think this is OK for now. We should discuss how we should handle kernel "warmups" throughout RAJA Perf and apply an approach uniformly throughout.

src/apps/INTSC_HEXHEX-Hip.cpp
```diff
@@ -0,0 +1,215 @@
...
+    stopTimer();
+
+  } else if ( vid == Lambda_HIP ) {
+
+    startTimer();
+    for (RepIndex_type irep = 0; irep < run_reps; ++irep) {
+
+      const Size_type grid_size = RAJA_DIVIDE_CEILING_INT(iend, block_size);
+
+      auto intsc_hexhex_lambda = [=] __device__
```
rhornung67: Please see my earlier suggestion for formatting lambda definition code.

src/apps/INTSC_HEXHEX-Hip.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+           Index_type blksize   = block_size ;
+           Index_type blk       = i / block_size ;
+           Index_type ith       = i ;
+           Index_type thridx    = i % block_size ;
+
+           Real_ptr vv_int_p = (Real_ptr ) vv_int + 8*blk ;
+           INTSC_HEXHEX_BODY; };
+
+      auto intsc_hexhex_fixup_lambda = [=] __device__
```
rhornung67: Please see my earlier suggestion for formatting lambda definition code.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,277 @@
...
+    Real_type v0, vx, vy, vz ;
+
+    Real_type xmin = m_xmin, ymin = m_ymin, zmin = m_zmin ;
+    Real_type xmax = m_xmax, ymax = m_ymax, zmax = m_zmax ;
+
+    if ( m_shift > 0.0 ) {
+      xmin += m_shift ;   ymin += m_shift ;   zmin += m_shift ;
+    } else {
+      xmax -= m_shift ;   ymax -= m_shift ;   zmax -= m_shift ;
+    }
```
rhornung67, line 176:
```suggestion
    }
    
```

Blank lines between code chunks helps readability.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,277 @@
...
+    Real_type dx = xmax - xmin, dy = ymax - ymin, dz = zmax - zmin ;
```
rhornung67:
```suggestion
    Real_type dx = xmax - xmin;
    Real_type dy = ymax - ymin;
    Real_type dz = zmax - zmin ;
    
```

This is more readable.

src/apps/INTSC_HEXHEX.cpp
```diff
@@ -0,0 +1,277 @@
...
+           ( dym*dym > tolsqy ) ||
+           ( dzm*dzm > tolsqz ) ) {
+
+        Size_type paragraph_max = 1000 ;
+
+        std::string sbuf  ;
+        std::vector<Char_type> vbuf(paragraph_max) ;
+        Char_ptr buf = vbuf.data() ;
+
+        snprintf ( buf, paragraph_max, "%s %s %s.\n",
```
rhornung67: Please use C++ I/O operations and the "getCout" method we have in the common directory. The method we have will not print multiple copies of output when running with multiple MPI ranks.

Nevermind. I see that you are writing to a buffer and sending that to getCout().  Please disregard.

src/apps/INTSC_HEXRECT.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include "common/KernelBase.hpp"
+
+namespace rajaperf
+{
+class RunParams;
+
+namespace apps
+{
+class ADomain;
```
rhornung67: Is this forward declaration needed? It doesn't look like it's used anyplace.

src/apps/INTSC_HEXHEX.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include "common/KernelBase.hpp"
+
+namespace rajaperf
+{
+class RunParams;
+
+namespace apps
+{
+class ADomain;
```
rhornung67: s this forward declaration needed? It doesn't look like it's used anyplace.

src/apps/INTSC_HEXRECT.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/2c16aa6eb0096d8df68209c0d07a52c97539d336

src/apps/INTSC_HEXHEX.hpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/2c16aa6eb0096d8df68209c0d07a52c97539d336

src/apps/INTSC_HEXHEX-Cuda.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/1e9b2794ad78a35997fc9037a1bffe81e7fe2d8a

src/apps/INTSC_HEXHEX-Cuda.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/1e9b2794ad78a35997fc9037a1bffe81e7fe2d8a

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/1e9b2794ad78a35997fc9037a1bffe81e7fe2d8a

src/apps/INTSC_HEXHEX-Hip.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/1e9b2794ad78a35997fc9037a1bffe81e7fe2d8a

src/apps/INTSC_HEXHEX.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/29e178fd99e60618974073363d538008f568d5b2

jmgrandy: (I didn't see your second comment.  I can revert this if you prefer).

src/apps/INTSC_HEXHEX.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/701606dafd8367bdd10cd48e0fd41f8360d10c17

src/apps/INTSC_HEXHEX.cpp
jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/701606dafd8367bdd10cd48e0fd41f8360d10c17

src/apps/INTSC_HEXHEX.cpp
MrBurmark: Nice, how about just using getCout() instead of a new ostringstream?

jmgrandy: https://github.com/LLNL/RAJAPerf/pull/554/commits/2cfdb9ac2c83b55c490d67bb85395cd084b98041

PR #557, opened 2025-09-26
Combine tsteps loops into repetition loops

MrBurmark:
# Summary
Combine the loops over tsteps found in a few polybench kernels with their repetition loops. This makes these kernels more similar with other kernels in the perf suite where the fundamental algorithm is done once per repetition instead of many times.

- This PR is a refactoring
- It does the following:
  - Modifies/refactors Polybench ADI, FDTD_2D, HEAT_3D, JACOBI_1D, and JACOBI_2D to combine the tsteps and repetition loops.

# Discussion
For ADI and FDTD_2D there is a dependency on the number of tsteps and/or the actual timestep. We should discuss how to handle that. For now I have done things in a way that preserves the previous answers.


rhornung67 (approved):
We can mess with ADI and FDTD_2D to remove the tsteps loops. We just have to make sure that compilers don't optimize anything away.

src/polybench/POLYBENCH_FDTD_2D.hpp
```diff
@@ -9,28 +9,28 @@
 ///
 /// POLYBENCH_FDTD_2D kernel reference implementation:
 ///
-/// for (t = 0; t < TSTEPS; t++)
-/// {
+/// // removed loop [0, TSTEPS)
+///
+/// for (j = 0; j < ny; j++) {
+///   ey[0][j] = fict[t];
+/// }
```
MrBurmark, line 16: Here is where FDTD_2D depends on the tstep.

src/polybench/POLYBENCH_FDTD_2D-Seq.cpp
```diff
@@ -31,29 +31,26 @@ void POLYBENCH_FDTD_2D::runSeqVariant(VariantID vid, size_t RAJAPERF_UNUSED_ARG(
...
+        }
+        for (Index_type i = 0; i < nx - 1; i++) {
+          for (Index_type j = 0; j < ny - 1; j++) {
+            POLYBENCH_FDTD_2D_BODY4;
           }
+        }
 
-        }  // tstep loop
-
+        t = (t+1) % m_tsteps;
```
MrBurmark, line 53: Here is where I calculate which tstep we would have been on.

pearce8 (approved):

MrBurmark:
@rhornung67 the gitlab checks don't seem to be working for this PR.

PR #589, opened 2025-11-25
Add more variants for Massvec3D

artv3:
DRAFT-PR. -- This MR adds the option to store threadblock info in the launch ctx avoiding calling blockDim.x,y,z during the loop methods in raja launch

### DIFF
--- a/src/apps/INTSC_HEXHEX.hpp
+++ b/src/apps/INTSC_HEXHEX.hpp
@@ -155,6 +155,10 @@ public:
 
   void runSeqVariant(VariantID vid, size_t tune_idx);
   void runOpenMPVariant(VariantID vid, size_t tune_idx);
+  void runOpenMPTargetVariant(VariantID vid, size_t)
+  {
+    getCout() << "\n INTSC_HEXHEX : Unknown OMP Target variant id = " << vid << std::endl;
+  }
   void runCudaVariant(VariantID vid, size_t tune_idx);
   void runHipVariant(VariantID vid, size_t tune_idx);
 
--- a/src/apps/INTSC_HEXRECT.cpp
+++ b/src/apps/INTSC_HEXRECT.cpp
@@ -336,8 +336,8 @@ void INTSC_HEXRECT::setUp(VariantID vid,
 
   // which zones to intersect.  Computed by hand for this test of
   // the geometry kernel.
-  allocDataForInit ( m_intsc_d, m_nrecords, vid ) ;
-  allocDataForInit ( m_intsc_t, m_nrecords, vid ) ;
+  auto a_id = allocDataForInit ( m_intsc_d, m_nrecords, vid ) ;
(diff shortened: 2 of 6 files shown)


src/apps/MASSVEC3DPA-Hip.cpp
```diff
@@ -491,10 +571,171 @@ void MASSVEC3DPA::runHipVariantImpl(VariantID vid, size_t tune_idx)
...
+      using inner_x = RAJA::LoopPolicy<RAJA::hip_ctx_thread_loop_x>;
+
+      using inner_y = RAJA::LoopPolicy<RAJA::hip_ctx_thread_loop_y>;
+
+      using inner_z = RAJA::LoopPolicy<RAJA::hip_ctx_thread_loop_z>;
+
+      startTimer();
+      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
+
+        RAJA::launch<launch_policy>(
```
artv3: TODO: move the RAJA code into a function that is templated on policies to avoid duplicated code...

src/apps/MASSVEC3DPA-Hip.cpp
```diff
@@ -135,6 +135,72 @@ void MassVec3DPA_ARGUMENT_LOOP_INC(const Real_ptr B, const Real_ptr Bt,
...
+                              const Real_ptr D, const Real_ptr X,
+                              Real_ptr Y,
+                              const Index_type runtime_block_size)
+{
+
+  const Index_type e = blockIdx.x;
+
+  const int bdx = blockDim.x; //block size is the same for xyz...
+  //const int bdy = blockDim.y;
+  //const int bdz = blockDim.z;
```
artv3: Is it worth adding the different dimensions to match the RAJA variant?

MrBurmark: Matching is nice.

MrBurmark: What about the argument runtime_block_size above though? Should that be removed?

src/apps/MASSVEC3DPA-Hip.cpp
MrBurmark: How much of the body could be a macro?

src/apps/MASSVEC3DPA-Hip.cpp
```diff
@@ -249,6 +324,166 @@ void MassVec3DPA_DIRECT(const Real_ptr B, const Real_ptr Bt,
   } // (c) dimension loop
 }
 
+template<typename inner_x, typename inner_y, typename inner_z, typename RESOURCE>
+void MASSVEC3DPA::runRAJAImpl(RESOURCE &res)
```
artv3, line 328: True to the RAJA mantra, "Change the policy not src code!"

artv3:
TODO: remove Bt as it is not used.

PR #592, opened 2025-12-04
Quiet C++20 compiler warnings

rhornung67:
# Summary

- This PR includes a code change to quiet a compiler warning for C++20 builds that is reported for every loop over kernel repetitions caused by our use a a volatile qualified loop counter variable, which we do to prevent unwanted compiler optimizations in the context of the loop around the actual kernels.
- Specifically, many compilers emit a deprecation warning when using a volatile qualified counter variable as a counter in a for-loop even when the value is discarded in the for-loop’s third expression. In practice, there is no way to keep a volatile loop counter in the for update slot without triggering the warning in GCC or Clang.
- This code change is a way to eliminate the warning and seems reasonably no-disruptive. Please note that there are a couple of minor variations that are needed to quiet the warnings in different situations.


src/basic/DAXPY-Seq.cpp
```diff
@@ -37,7 +37,10 @@ void DAXPY::runSeqVariant(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
     case Base_Seq : {
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
+      // Standalone assignment expression for loop counter to quiet
+      // C++20 compiler warning 
+      for (RepIndex_type irep = 0; irep < run_reps; /* no update here */) {
+        irep = irep + 1;
```
MrBurmark: This seems ok if it quiets the warning.

MrBurmark: Would it make any sense to try to hide this whole thing in a macro?

rhornung67: I think that would be more opaque and awkward.

rhornung67:
@adayton1 not sure how that applies to our case. The loop itself does not get optimized away, rather operation inside it may not get done on every loop iteration.  The awkward syntax here is to prevent compiler warning spew for something that is deprecated in C++20.

rchen20 (approved):

src/algorithm/ATOMIC-Seq.cpp
```diff
@@ -52,7 +53,8 @@ void ATOMIC::runSeqVariantReplicate(VariantID vid)
                                };
 
       startTimer();
-      for (RepIndex_type irep = 0; irep < run_reps; irep = irep + 1) {
+      // Awkward expression for loop counter quiets C++20 compiler warning
+      for (RepIndex_type irep = 0; irep < run_reps; static_cast<void>(((irep = irep + 1), 0))) {
```
adayton1, line 57: Why do some files use static_cast<void> while others don't?

PR #599, opened 2025-12-10
Change FEMSWEEP to read in file for mesh data

rchen20:
# Summary

- This PR is a refactoring
- It does the following:
  - Refactors FEMSWEEP to read in a file to populate mesh connectivity data.

### DIFF
--- a/src/common/RunParams.cpp
+++ b/src/common/RunParams.cpp
@@ -407,6 +407,14 @@ void RunParams::parseCommandLineOptions(int argc, char** argv)
       i++;
       if ( i < argc ) {
         npasses = ::atoi( argv[i] );
+        #if defined(RAJA_PERFSUITE_USE_CALIPER)
+          if (npasses > 1) {
+            getCout() << "\nBad input:"
+                  << " --npasses cannot be >1 when profiling with Caliper"
+                  << std::endl;
+            input_state = BadInput;
+          }
+        #endif
       } else {
         getCout() << "\nBad input:"
                   << " must give --npasses a value for number of passes (int)"
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,12 +1,23 @@
 Intellectual Property Notice
 ------------------------------
 
+RAJA Performance Suite is produced at the Lawrence Livermore National Laboratory.
+
+Unlimited Open Source - BSD Distribution
+LLNL-CODE-738930
+OCEC-17-159
+
+This file is part of RAJA Performance Suite. 
+For details, see https://github.com/llnl/RAJAPerf
+
 RAJA Performance Suite is licensed under the BSD 3-Clause license
 (BSD-3-Clause or https://opensource.org/licenses/BSD-3-Clause).
 
-Copyrights and patents in the RAJA Performance Suite  project are retained
+Copyrights and patents in the RAJA Performance Suite project are retained
 by contributors. No copyright assignment is required to contribute to the
 RAJA Performance Suite.
+
+See https://github.com/llnl/RAJAPerf/contributors for the list of contributors.
 
 Additional BSD Notice
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,18 +1,7 @@
+BSD 3-Clause License
+
 Copyright (c) 2017-2026, Lawrence Livermore National Security, LLC.
-
-Produced at the Lawrence Livermore National Laboratory
-
-Unlimited Open Source - BSD Distribution
-LLNL-CODE-738930
-OCEC-17-159
-
 All rights reserved.
-
-This file is part of RAJA Performance Suite
-
-For details, see https://github.com/llnl/RAJAPerf
-
-Please also read RAJAPerf/COPYRIGHT
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:


src/apps/FEMSWEEP.cpp
```diff
@@ -98,6 +99,67 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
   allocAndInitDataRandValue (m_M0dat    , m_M0len     , vid);
   allocAndInitDataRandValue (m_Xdat     , m_Xlen      , vid);
 
+  // Read mesh connectivity data from file.
```
MrBurmark, line 221: This should probably use allocForInit instead of copying the data through an extra buffer. See an example here. https://github.com/LLNL/RAJAPerf/blob/06fc079971f6491f1ed9cee3369ef214527bb84f/src/algorithm/HISTOGRAM.cpp#L57

MrBurmark:
Does reading the data also affect the problem size, number of elements, etc?

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
+      }
+    }
+
+    else if ( line == std::string("m_nz") )
+    {
+      // Read next line for value.
+      if ( std::getline(dataFile, line) )
+      {
+        m_nz = std::stoi(line);
+        //std::cout << "RCC CHECK m_nz " << m_nz << std::endl;
```
artv3: remove commented code?

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
+      }
+    }
+
+    else if ( line == std::string("m_na") )
+    {
+      // Read next line for value.
+      if ( std::getline(dataFile, line) )
+      {
+        m_na = std::stoi(line);
+        //std::cout << "RCC CHECK m_na " << m_na << std::endl;
```
artv3: Remove commented code?

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
+      }
+    }
+
+    else if ( line == std::string("m_ng") )
+    {
+      // Read next line for value.
+      if ( std::getline(dataFile, line) )
+      {
+        m_ng = std::stoi(line);
+        //std::cout << "RCC CHECK m_ng " << m_ng << std::endl;
```
artv3: Remove commented code?

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
+        std::cout << "Unable to initialize m_ng properly in constructor. Please check the " << mesh_file << " file." << std::endl;
+      }
+    }
+
+  }
+
   m_ne = m_nx * m_ny * m_nz;
-  m_na = 72;
-  m_ng = 128;
+  //std::cout << "RCC CHECK m_ne " << m_ne << std::endl;
```
artv3: Remove commented code?

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
 
   setDefaultProblemSize(ND * m_ne * m_ng * m_na);
   setDefaultReps(1);
 
+  m_sharedinteriorfaces = (m_nx - 1) * m_ny * m_nz +
+                          m_nx * (m_ny - 1) * m_nz +
+                          m_nx * m_ny * (m_nz - 1);
+  //std::cout << "RCC CHECK m_sharedinteriorfaces " << m_sharedinteriorfaces << std::endl;
```
artv3:
```suggestion
```

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
+  m_boundaryfaces = 2 * m_nx * m_ny + 2 * m_ny * m_nz + 2 * m_nx * m_nz;
+  //std::cout << "RCC CHECK m_boundaryfaces " << m_boundaryfaces << std::endl;
```
artv3:
```suggestion
```

src/apps/FEMSWEEP.cpp
```diff
@@ -26,20 +27,110 @@ namespace apps
...
+  m_hplanes = m_nx + m_ny + m_nz - 2;
+  //std::cout << "RCC CHECK m_hplanes " << m_hplanes << std::endl;
```
artv3:
```suggestion
```

src/common/RunParams.cpp
```diff
@@ -713,7 +716,7 @@ void RunParams::parseCommandLineOptions(int argc, char** argv)
 
     } else if ( opt == std::string("--gpu_block_size") ) {
 
-      bool got_someting = false;
+      bool got_something = false;
```
artv3, line 785: got what?

src/common/RunParams.cpp
```diff
@@ -959,7 +975,7 @@ void RunParams::parseCommandLineOptions(int argc, char** argv)
                 opt == std::string("--sycl-mpi-data-space") ||
                 opt == std::string("--kokkos-mpi-data-space") ) {
 
-      bool got_someting = false;
+      bool got_something = false;
```
artv3, line 1044: Not sure I understand what the boolean mean

src/apps/FEMSWEEP.cpp
artv3:
```suggestion
```

src/apps/FEMSWEEP.cpp
artv3:
```suggestion
```

src/apps/FEMSWEEP.cpp
artv3:
```suggestion
```

src/common/RunParams.cpp
MrBurmark: A block size, but this is just fixing spelling so we can worry about making the naming better some other time.

src/common/RunParams.cpp
```diff
@@ -930,6 +933,19 @@ void RunParams::parseCommandLineOptions(int argc, char** argv)
...
+
+      i++;
+      if ( i < argc ) {
+        opt = std::string(argv[i]);
+        if ( opt.at(0) == '-' ) {
+          i--;
+        } else {
+          femsweep_mesh_file = std::string( argv[i] );
+        }
+      }
```
MrBurmark, line 1013: Can we exclude FEMSWEEP by default if not given a file?

artv3: second this

src/common/RunParams.cpp
rchen20: Just a spelling typo that's not related to the femsweep.

src/apps/FEMSWEEP.cpp
```diff
@@ -90,18 +172,319 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
...
+  // Read mesh connectivity data from file.
+  std::ifstream dataFile(mesh_file);
+
+  if ( !dataFile.is_open() )
+  {
+    std::cout << "Could not open " << mesh_file << " in setUp." << std::endl;
+  }
+
+  int lcount = 0;
+  int sizetemp = 0;
```
MrBurmark: If these are not used across different scopes they should be localized.

src/apps/FEMSWEEP.cpp
```diff
@@ -90,18 +172,319 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
...
+        sizetemp = std::stoi(line);
+        // Check size for sanity.
+        if ( sizetemp != m_na )
+        {
+          std::cout << "Size of m_nhpaa_r in " << mesh_file << " does not match." << std::endl;
+        }
+        auto temp_nhpaa_r = allocDataForInit(m_nhpaa_r, sizetemp, vid); 
+        // Read rest of entries for array.
+        lcount = 0;
+        while ( lcount < sizetemp )
```
MrBurmark: Should this be a for loop?

src/apps/FEMSWEEP.cpp
```diff
@@ -90,18 +172,299 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
...
+  if ( !dataFile.is_open() )
+  {
+    std::cout << "Could not open " << mesh_file << " in setUp." << std::endl;
+  }
+
+  std::string line;
+  while ( std::getline(dataFile, line) )
+  {
+
+    if ( line == std::string("m_nhpaa_r") )
```
MrBurmark:
```suggestion
    if ( line == "m_nhpaa_r" )
```
No need to cast to string if the other side is a string.

src/apps/FEMSWEEP.cpp
```diff
@@ -90,18 +172,299 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
...
+    {
+      // Read next line for array size.
+      if ( std::getline(dataFile, line) )
+      {
+        int sizetemp = std::stoi(line);
```
MrBurmark: Are these all ints, would it make sense to read them as a wider data type?

src/apps/FEMSWEEP.cpp
```diff
@@ -90,18 +172,299 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
...
+  }
+
+  std::string line;
+  while ( std::getline(dataFile, line) )
+  {
+
+    if ( line == std::string("m_nhpaa_r") )
+    {
+      // Read next line for array size.
+      if ( std::getline(dataFile, line) )
```
MrBurmark: This pattern of reading in data from a file to fill an array happens multiple times, make a function?

src/apps/FEMSWEEP.cpp
rchen20: Yes, they should all be ints for the mesh data. Any doubles are generated randomly or set to 0.0 before the actual sweep.

MrBurmark: I meant should they be longs as Index_type is a 64-bit type.

src/apps/FEMSWEEP.hpp
```diff
@@ -307,31 +310,26 @@ class FEMSWEEP : public KernelBase
   using gpu_block_sizes_type = integer::make_gpu_block_size_list_type<default_gpu_block_size,
                                                          integer::MultipleOf<32>>;
 
+  std::string mesh_file;  // Path to file name of femsweep mesh.
```
artv3: I think that should be m_mesh_file to be consistent with the rest of the member variables

src/apps/FEMSWEEP.cpp
```diff
@@ -88,6 +170,40 @@ FEMSWEEP::~FEMSWEEP()
 {
 }
 
+template < typename T >
+void FEMSWEEP::readIndexArray(VariantID vid, std::ifstream & file, T*& arr, Index_type expectedsize, std::string arrname)
```
MrBurmark, line 179: This can probably be a free function, probably in an unnamed namespace.

src/apps/FEMSWEEP.cpp
```diff
@@ -26,12 +27,88 @@ namespace apps
...
+  }
+
+  std::string line;
+  while ( std::getline(dataFile, line) )
+  {
+
+    if ( line == std::string("m_nx") )
+    {
+      // Read next line for value.
+      if ( std::getline(dataFile, line) )
```
MrBurmark: How about making a function for reading these too?

MrBurmark: Or use a lambda?

MrBurmark (approved):

src/apps/FEMSWEEP.cpp
rchen20: I can move it if some other kernel is going to use it, but I think the femsweep is the only one that requires something like this at the moment so I'll keep it the way it is.

MrBurmark: I just mean that it doesn't use any of the kernel's state, it could be a free function. Its not a big deal, but its probably best practice if that's true to make it a free function.

rchen20:
@rhornung67 How about I revert this PR such that there is only one small mesh? That way the meshes won't show up in the git branches and history. After that, we can create another repo to store the larger meshes.

artv3:
> > @rchen20 , you can also squash merge
> 
> Thanks, but I don't think squash merging will prevent people from downloading all the meshes on every `git clone`. I'd need to remove the meshes from the history entirely.

Ah I see, thanks for explaining. Sure I'm in favor of your original proposition

PR #600, opened 2025-12-10
Change rep loop increment to a macro

rhornung67:
The macro quiets C++20  warnings for good this time!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,6 +12,7 @@ add_subdirectory(common)
 add_subdirectory(basic)
 add_subdirectory(basic-kokkos)
 add_subdirectory(apps)
+add_subdirectory(apps-kokkos)
 add_subdirectory(lcals)
 add_subdirectory(lcals-kokkos)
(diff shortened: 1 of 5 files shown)


src/algorithm/ATOMIC-OMP.cpp
```diff
@@ -34,7 +34,6 @@ void ATOMIC::runOpenMPVariantReplicate(VariantID vid)
     case Base_OpenMP : {
 
       startTimer();
-      // Awkward expression for loop counter quiets C++20 compiler warning
       for (RepIndex_type irep = 0; irep < run_reps; static_cast<void>(((irep = irep + 1), 0))) {
```
MrBurmark: Why weren't these changed to use the macro?

rhornung67: My Python script didn't look for those. Me culpa. I fixed these.

MrBurmark (approved):

adayton1 (approved):

PR #605, opened 2025-12-17
Output Amount of Memory Touched

MrBurmark:
# Summary

Add a count for the amount of memory touched. This is useful as a estimate of the memory used by a kernel and a best case for the amount of cache used by a kernel.

This is implemented by counting the bytes modify written separately from bytes read and bytes written.

- This PR is a feature
- It does the following:
  - Adds more accurate memory usage at the request of people trying to run problems at 4x the last level cache size

### DIFF
--- a/docs/sphinx/user_guide/run.rst
+++ b/docs/sphinx/user_guide/run.rst
@@ -51,29 +51,12 @@ kernel.
 Running the Suite
 ==================
 
-After compilation, the main executable will reside in the ``bin`` subdirectory 
-of the build space. The executable will be able to run all kernels and 
-variants that have been built depending on which CMake options were specified
-to configure the build.
+After compilation, the RAJA Performance Suite executable will reside in the
+``bin`` subdirectory of the build space. The executable will be able to run
+all kernels and variants that have been built depending on which CMake options
+were specified to configure the build.
 
-To run the Suite in its default mode, type the executable name with no 
-command-line arguments::
-
-  $ ./bin/raja-perf.exe
-
-This will run all kernels and variants that have been built in their default
-configurations. Information describing how the Suite will run along with
-some information about each kernel will appear on the screen. More information
-about kernel and execution details will also appear in a run report files 
-generated in the run directory after Suite execution completes. 
-
-.. note:: You can pass the ``--dryrun`` command-line option to the executable
-          to see a summary of how the Suite will execute, by showing default
-          run parameters, without actually running it. You can also pass 
-          other command-line options when doing a "dry run" and you will see
-          that the given options are represented in the screen output.
-
-The Suite can be run in a variety of ways determined by the command-line 
+The Suite can be run in many different ways chosen by the command-line 
 options passed to the executable. For example, you can run or exclude subsets 
 of kernels, variants, or groups. You can also pass options to set problem 
 sizes, number of times each kernel is run (sampled), and many other run 
@@ -94,6 +77,27 @@ or::
 
 .. note:: To see all available Suite execution options, pass the `--help` or 
           `-h` option to the executable.
+
+.. important:: We do not describe most of the Suite execution options in this
+               guide since the runtime help output is the main reference for
+               available options, defaults, and arguments they accept.
+
+To run the Suite in its default mode, type the executable name with no 
+command-line arguments::
+
+  $ ./bin/raja-perf.exe
+
+This will run all kernels and variants that have been built in their default
+configurations. Information describing how the Suite will run along with
+some information about each kernel will appear on the screen. More information
+about kernel and execution details will also appear in a run report files 
+generated in the run directory after Suite execution completes. 
+
+.. note:: You can pass the ``--dryrun`` command-line option to the executable
+          to see a summary of how the Suite will execute by showing run
+          parameters without actually running it. You can pass any other
+          command-line options when doing a "dry run" and you will see
+          that the given options are represented in the screen output.
 
 Lastly, the program will report specific errors if given incorrect input, such
 as an option that requires a value and no value is provided. It will also emit 
@@ -116,7 +120,8 @@ will report the following in the screen output::
     See run parameters or option messages above.
 
 The output indicates that the kernel input is invalid because the string Foo
-is not the name of a kernel in the Suite, while DAXPY is the name of a kernel. 
+is not the name of a kernel in the Suite, while DAXPY is the name of a kernel
+in the Suite.
 
 .. note:: The Suite executable will attempt to provide helpful information
           if it is given incorrect input, such as command-line arguments that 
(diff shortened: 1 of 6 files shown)


docs/sphinx/dev_guide/kernel_class_impl.rst
```diff
@@ -60,6 +65,17 @@ The methods in the source file are:
         not every kernel implements every variant, so ``KernelBase`` provides a
         "default" implementation that defines no variants or tunings.
 
+    ..note:: When counting the number of bytes accessed each byte accessed is
+             counted in exactly one of the byte counters. For example a byte
```
rhornung67:
```suggestion
    ..note:: When counting the number of bytes accessed, each access is
             counted in exactly one of the counters. For example a byte
```

docs/sphinx/dev_guide/kernel_class_impl.rst
```diff
@@ -60,6 +65,17 @@ The methods in the source file are:
...
+             that read and written is counted in the bytes read, modified, and
+             written counter and not in the read or written counters. Each byte
```
rhornung67:
```suggestion
             that is read and written is counted in the bytes read, modified, and
             written counter and not in the read or written counters. Each byte
```

This is confusing. If I read and write some chunk of bytes, why would it not be counted in the read counter and in the write counter?  I guess what is represented by the different counters is not obvious or clear, at least to me.

docs/sphinx/dev_guide/kernel_class_impl.rst
```diff
@@ -60,6 +65,17 @@ The methods in the source file are:
...
+             accessed is counted once for each loop/kernel launch it is accessed
+             in even if it is accessed multiple times within a loop/kernel
+             launch. For example if a byte is written five times in the first
+             loop/kernel launch and read two times in the second loop/kernel
+             launch of a kernel, then it is counted once in the bytes read
```
rhornung67: When you say counted `once`, does that mean the byte count is one regardless if it is read/written multiple times?

MrBurmark: Yes, if the same byte is accessed multiple times in a kernel its only counted once.
The idea is that overall performance should be mostly affected by traffic to the memory and that reuse is cached.

docs/sphinx/dev_guide/kernel_class_impl.rst
MrBurmark: Perhaps on the user output side we could include the RMW bytes with the R bytes and the W bytes, but on the input side its easier to separate things into mutually exclusive categories, only have to list it once instead of 1 times.

MrBurmark: I'll add the RMW bytes into the R and W bytes for user outputs. But I think its easier to count them into mutually exclusive buckets when developing/adding a kernel..

docs/sphinx/dev_guide/kernel_class_impl.rst
MrBurmark: It could be useful to have a full accounting of the number of loads and stores or the amount of memory requests in total that we expect to hit the cache so we can have an idea of the amount of cache bandwidth we expect to use. However that is harder to count unless the code is manually optimized to avoid redundant reads/writes.

MrBurmark: I changed all of this text, please take another look @rhornung67.

docs/sphinx/user_guide/output.rst
```diff
@@ -80,21 +80,33 @@ Information reported in the file for each kernel is:
...
+    to and from memory assuming perfect cache reuse and ignoring partial usage
+    of data in some memory transactions.
   * **FLOPs/rep** -- Total number of floating point operations executed for 
     each repetition of kernel. Currently, we count arithmetic operations 
     (+, -, *, /) and functions, such as exp, sin, etc. as one FLOP. We do not 
     currently count operations like abs and comparisons (<, >, etc.) in the 
     FLOP count. So these numbers are rough estimates. For actual FLOP counts, 
     a performance analysis tool should be used.
+  * **BytesTouched/rep** -- Total number of bytes accessed for each repetition
+    of kernel. This is a best case scenario for the amount of cache needed to
```
artv3: byte in DRAM or HBM right?

docs/sphinx/user_guide/output.rst
```diff
@@ -80,21 +80,33 @@ Information reported in the file for each kernel is:
...
     each repetition of kernel.
+  * **BytesModifyWritten/rep** -- Total number of bytes modified for each
+    repetition of kernel. The intersection of bytes in both ``BytesRead/rep``
+    and ``BytesWritten/rep``.
   * **BytesAtomicModifyWritten/rep** -- Total number of bytes modified by 
     atomic operations in a kernel. If a kernel contains no atomic operations, 
     the value of zero is reported.
 
+  ..note:: The Bytes*/rep and FLOPs/rep counts are estimates for kernels
+           involving randomness or difficult to count algorithms.
```
artv3, line 113: Which kernels exhibit randomness? It may be nice to list the kernels which do not have the best estimates

rhornung67: @artv3 I think the plan is to have that sort of information appear in the output files. Putting kernel-specific information in the user documentation creates extra maintenance burden.

docs/sphinx/user_guide/output.rst
MrBurmark: Yes, each byte is counted once even if its accessed multiple times or in multiple ways. This doesn't count things we expect to be on the stack however like what is captured in the lambda.

docs/sphinx/user_guide/output.rst
MrBurmark: There is not currently an attribute for that, but it can be added fairly easily.

docs/sphinx/user_guide/output.rst
MrBurmark: @artv3 do you think this text needs clarification?

docs/sphinx/user_guide/output.rst
MrBurmark: I'll do that in a different PR, see #611.

docs/sphinx/user_guide/output.rst
```diff
@@ -80,21 +80,33 @@ Information reported in the file for each kernel is:
...
+    each repetition of kernel. This is a best case scenario of the total traffic
+    to and from memory assuming perfect cache reuse and ignoring partial usage
+    of data in some memory transactions.
   * **FLOPs/rep** -- Total number of floating point operations executed for 
     each repetition of kernel. Currently, we count arithmetic operations 
     (+, -, *, /) and functions, such as exp, sin, etc. as one FLOP. We do not 
     currently count operations like abs and comparisons (<, >, etc.) in the 
     FLOP count. So these numbers are rough estimates. For actual FLOP counts, 
     a performance analysis tool should be used.
+  * **BytesTouched/rep** -- Total number of bytes accessed for each repetition
```
artv3:
```suggestion
  * **BytesTouched/rep** -- Total number of bytes accessed in DRAM or HBM for each repetition
```

docs/sphinx/user_guide/output.rst
artv3: you had it listed somewhere else, but yeah I think adding it in would be nice. See suggestion

docs/sphinx/user_guide/output.rst
MrBurmark: There are some other attributes with string instead of integer values like the complexity of the kernel that should be printed with the other kernel info but are not, I will address that in a future PR as well.

docs/sphinx/user_guide/output.rst
MrBurmark: What about "memory" to be more general, or is that confusing because of the cache?

artv3: Yeah I thought it would be confusing because of the cache , so clarifying the memory I thought would be good

MrBurmark: Given that there are multiple Bytes attributes, would it be better to write this in a note, in the first "Bytes" attribute description, or in each "bytes" description?

artv3: thats a good point, yeah a note works

artv3 (approved):

PR #606, opened 2025-12-17
Rework how warmup kernels are specified and run

rhornung67:
# Summary

- This PR makes it easier to specify and run warmup kernels in different ways (turn-off warmups, our default feature-based warmup strategy, specify a specific set of kernels to run for warmup, or warmup all kernels that will be run when the suite runs).
- Note that this PR introduces a simple, but admittedly unseemly change involving a static variable in the `KernelBase` class to make warmup kernels run only once regardless of the specified # run reps or kernel default run reps.  We can change this later to something more desirable with more  involved code changes.
- Also, fixed some output file table header names.

### DIFF
--- a/docs/sphinx/user_guide/run.rst
+++ b/docs/sphinx/user_guide/run.rst
@@ -141,9 +141,8 @@ the same code for each kernel on each rank with minimal synchronization points
 to gather execution timing data from all ranks. This capability is provided so
 that individual kernel performance more closely aligns with how such kernels 
 would perform in a real application. For example, compute node memory bandwidth
-impact on performance may be different when running on many core system using
-OpenMP multithreading to exercise all cores than when each core is mapped to
-an MPI rank.
+may be different when running on a many core system using OpenMP multithreading
+to exercise all cores than when each core is mapped to an MPI rank.
 
 Running the Suite on multiple MPI ranks is just like running any other MPI
 application. For example, issuing the following command on a machine with


MrBurmark (approved):

src/common/RunParams.cpp
```diff
@@ -1348,11 +1354,16 @@ void RunParams::printHelpMessage(std::ostream& str) const
...
-      << "\t ========================================\n\n";;
+      << "\t ========================================\n\n";
+
+  str << "\t For warmup kernels, the default (no option specified) is to run a minimal set of warmup kernels based on\n"
+      << "\t RAJA features exercised in kernels selected to run. Other options are:\n\n";
+
+  str << "\t --warmup-disable (do not run any warmup kernels)\n\n";
 
-  str << "\t --disable-warmup (disable warmup kernels) [Default is run warmup kernels that are relevant to kernels selected to run]\n\n";
+  str << "\t --warmup-to-run (run each kernel specified to run once in warmup pass)\n\n";
```
MrBurmark: I'm not quite sure this is clear as you don't always specify certain kernels to run and sometimes specify some not to run.
```suggestion
  str << "\t --warmup-to-run (run the same kernels that are being benchmarked in the warmup pass)\n\n";
```

src/common/RunParams.cpp
```diff
@@ -1140,9 +1142,13 @@ void RunParams::parseCommandLineOptions(int argc, char** argv)
         input_state = DryRun;
       }
 
-    } else if ( std::string(argv[i]) == std::string("--disable-warmup") ) {
+    } else if ( std::string(argv[i]) == std::string("--warmup-disable") ) {
+
+      warmup_mode = WarmupMode::Disable;
 
-      disable_warmup = true;
+    } else if ( std::string(argv[i]) == std::string("--warmup-to-run") ) {
```
MrBurmark: I'm not quite sure that this name is clear. I read it as "warmup to run ... what?"

MrBurmark: What about `--warmup-same` or `warmup-all`?

src/common/RunParams.cpp
rhornung67: I changed the option name and explanation. Hopefully, it's reasonably clear now.

src/common/RunParams.cpp
rhornung67: I changed the name.

src/common/RunParams.cpp
MrBurmark: Its more clear now

docs/sphinx/user_guide/run.rst
```diff
@@ -130,8 +135,19 @@ is not the name of a kernel in the Suite, while DAXPY is the name of a kernel.
...
-For example, issuing the following command on a machine with slurm scheduling::
+The Suite can be configured and compiled to run in a distributed memory
+parallel mode using MPI. Running the Suite on multiple MPI ranks will execute
+the same code for each kernel on each rank with minimal synchronization points
+to gather execution timing data from all ranks. This capability is provided so
+that individual kernel performance more closely aligns with how such kernels 
+would perform in a real application. For example, compute node memory bandwidth
+impact on performance may be different when running on many core system using
+OpenMP multithreading to exercise all cores than when each core is mapped to
+an MPI rank.
```
MrBurmark:
```suggestion
would perform in a real application. For example, compute node memory bandwidth
may be different when running on many core system using
OpenMP multithreading to exercise all cores than when each core is mapped to
an MPI rank.
```

rhornung67: Fixed.  @MrBurmark if you have no other suggestions, I will merge when checks are green.

Issue #604, opened 2025-12-16
Correctness/Robustness Checking


MrBurmark:
When running multiple tunings or adding new tunings the answers can be inconsistent if there is a platform, software, or logic bug. 

Dying on inconsistent checksums would help quickly identify these. We would need to separate out cases (like atomics) where checksums are not reproducible. 

It would be helpful to be able to distinguish cases where the checksum is always the same, differs due to operation order between tunings, or always differs because atomics are used.

rhornung67:
We do something like this in the test we have https://github.com/llnl/RAJAPerf/blob/develop/test/test-raja-perf-suite.cpp#L183 However, this can be improved. I just set one tolerance for all kernels. We could expand this to define a different tolerance for each kernel, each subset of kernels, etc.  Is this what you were thinking?

MrBurmark:
The PASS/FAIL is enough to easily see that a checksum has failed, no need for the program to die on a failure.

PR #607, opened 2025-12-18
Add ChecksumConsistency attribute

MrBurmark:
This records whether all variant tunings of a kernel always get the same answer, or if the answer varies between variant tunings but is consistent for each variant tuning, or if the answer varies from run to run (atomics).

This attribute is implemented like computational complexity.

I chose values that are consistent with the current tunings. However some kernels could be implemented with atomics which would move them to inconsistent. For example kernels with dot products which are currently implemented as sequential loops even on GPUs.

# Summary

- This PR is a feature
- It does the following:
  - Adds checksum consistency kernel attribute at the request of #604 though this is only implements part of what is mentioned in that issue.

### DIFF
--- a/src/apps/LTIMES.cpp
+++ b/src/apps/LTIMES.cpp
@@ -53,7 +53,7 @@ LTIMES::LTIMES(const RunParams& params)
               ( static_cast<Checksum_type>(getDefaultProblemSize()) /
                                            getActualProblemSize() );
 
-  setChecksumConsistency(ChecksumConsistency::ConsistentPerVariantTuning);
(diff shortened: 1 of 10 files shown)


rhornung67 (commented):
Looks reasonable. Thank you for adding documentation to the dev guide. We should make sure some of the explanation of what is computed/reported also appears in the user guide section that describes the output files.

The checksum consistency info is not reported, in the checksum output file or screen output. I think it should be. Otherwise, how is it helpful to users?

rhornung67 (approved):

PR #608, opened 2025-12-19
Correctness Checking

MrBurmark:
# Summary

Add more support for correctness checking.
Each kernel now can set its own tolerance and we print if the checksum met that tolerance in the show-progress screen output and the checksum output file.
This mechanism is now also used by the test executable.

How concerned should we be that checksums grow over multiple passes?

- This PR is a feature
  - Adds correctness checking at the request of #604 
  - This needs to update RAJA once https://github.com/llnl/RAJA/pull/1969 is merged

## Example Output

Here is an example of -sp screen output.
```
Run kernel -- Basic_MULTI_REDUCE
        Running Base_Seq variant
                Running      default tuning -- 2.71738e-05 sec. x 50 rep. PASSED checksum
        Running Lambda_Seq variant
                Running      default tuning -- 2.71464e-05 sec. x 50 rep. PASSED checksum
        Running RAJA_Seq variant
                Running      default tuning -- 2.71234e-05 sec. x 50 rep. PASSED checksum
        Running Base_OpenMP variant
                Running      default tuning -- 0.000128336 sec. x 50 rep. PASSED checksum
        Running Lambda_OpenMP variant
                Running      default tuning -- 0.000129231 sec. x 50 rep. PASSED checksum
        Running RAJA_OpenMP variant
                Running      default tuning -- 1.70671e-05 sec. x 50 rep. FAILED checksum
```

Here is an example of checksum.txt file output.
```
Kernel                     
........................................................
Variants                   Result  Tolerance                   Average Checksum            Max Checksum Diff           Checksum Diff StdDev        
                                                                                           (vs. first variant listed)                              
----------------------------------------------------------------------------------------
Basic_MULTI_REDUCE         
........................................................
Base_Seq-default           PASSED  9.9999999999999995475e-08   5865.2700279413772306       0.0000000000000000000       0.0000000000000000000       
Lambda_Seq-default         PASSED  9.9999999999999995475e-08   5865.2700279413772306       0.0000000000000000000       0.0000000000000000000       
RAJA_Seq-default           PASSED  9.9999999999999995475e-08   5865.2700279413772306       0.0000000000000000000       0.0000000000000000000       
Base_OpenMP-default        PASSED  9.9999999999999995475e-08   5865.2700279413772306       0.0000000000000000000       0.0000000000000000000       
Lambda_OpenMP-default      PASSED  9.9999999999999995475e-08   5865.2700279413772306       0.0000000000000000000       0.0000000000000000000       
RAJA_OpenMP-default        FAILED  9.9999999999999995475e-08   2918.7998532503377369       2946.4701746910394893       2.2204460492503130808e-16 

-------------------------------------------------------
```

### DIFF
--- a/src/common/KernelBase.cpp
+++ b/src/common/KernelBase.cpp
@@ -650,7 +650,7 @@ void KernelBase::setCaliperMgrVariantTuning(VariantID vid,
     std::string check_profile;
     // If both not empty
     if (!updatedSpotConfig.empty() && !updatedCaliConfig.empty()) {
-      check_profile = "spot(" + updatedSpotConfig + ")," + updatedCaliConfig;
+      check_profile = updatedCaliConfig + ",spot(" + updatedSpotConfig + ")";
     }
     else if (!updatedSpotConfig.empty()) {
       check_profile = "spot(" + updatedSpotConfig + ")";
@@ -676,24 +676,27 @@ void KernelBase::setCaliperMgrVariantTuning(VariantID vid,
     cali::ConfigManager m;
     mgr[vid][tstr] = m;
     std::string vstr = getVariantName(vid);
-    std::string profile;
+    std::string profile = "";
+    std::string sprofile = "";
+    std::string ccprofile = "";
+    if (!updatedCaliConfig.empty()) {
+      ccprofile += updatedCaliConfig + ",";
+    }
     // If --outfile not provided, give generic name
     if (outfile == "RAJAPerf") {
-      profile = "spot(output=" + vstr + "-" + tstr + ".cali";
+      sprofile = "spot(output=" + vstr + "-" + tstr + ".cali";
     }
     else {
       // Ensure cali files for each variant/tuning are not same file name
       if (num_variants_tunings > 1)
         throw std::runtime_error("Error: Cannot use '--outfile' with Caliper if running multiple variants/tunings. Must be running single variant & tuning.");
-      profile = "spot(output=" + outfile + ".cali";
+      sprofile = "spot(output=" + outfile + ".cali";
     }
     if(!updatedSpotConfig.empty()) {
-      profile += "," + updatedSpotConfig;
+      sprofile += "," + updatedSpotConfig;
     }
-    profile += ")";
-    if (!updatedCaliConfig.empty()) {
-      profile += "," + updatedCaliConfig;
-    }
+    sprofile += ")";
+    profile = ccprofile + sprofile;
     std::cout << "Profile: " << profile << std::endl;
     mgr[vid][tstr].add_option_spec(kernel_info_spec);
     mgr[vid][tstr].set_default_parameter("rajaperf_kernel_info", "true");


artv3:
Do you mean multiple passes over the suite? or within the number of times the kernel is invoked?

docs/sphinx/dev_guide/kernel_class_impl.rst
```diff
@@ -52,12 +52,24 @@ The methods in the source file are:
...
-        slightly, and ``Inconsistent`` where the checksum of a variant tuning
-        may vary slightly run to run.
+      * The consistency of the checksums of the kernel. If the kernel
+        always produces the same checksum value for all variant tunings then the
+        checksums are ``Consistent``. Most kernels get a different but consistent
+        checksum for each variant tuning so the checksums are
+        ``ConsistentPerVariantTuning``. On the other hand, some kernels have
+        variant tunings that get different checksums on each run of that variant
+        tuning, for example due to the ordering of floating-point atomic add
+        operations, so the checksums are ``Inconsistent``.
```
artv3, line 67: Are ``Inconsistent`` kernels expected to agree within some tolerance? I suspect the ``Inconsistency`` may grow per number of reps?

src/algorithm/ATOMIC.cpp
```diff
@@ -34,6 +34,7 @@ ATOMIC::ATOMIC(const RunParams& params)
   setFLOPsPerRep(getActualProblemSize());
 
   setChecksumConsistency(ChecksumConsistency::Inconsistent); // atomics
+  setChecksumTolerance(ChecksumTolerance::normal);
```
artv3, line 38: Are the ``ChecksumTolerance::normal`` values defined in the docs?

MrBurmark: Nope, its only in the code. It is 1e-7. It might not be a bad idea to output the tolerance in the checksum output file.

docs/sphinx/dev_guide/kernel_class_impl.rst
MrBurmark: It depends on the kernel, most kernels overwrite their outputs on each rep instead of continuing to accumulate.

src/algorithm/ATOMIC.cpp
artv3: Sounds like a great idea to me! I would also encourage documentation on what we can expect from normal vs tight tolerance. I haven't dug into it myself, but I'm curious if the tolerances are relative and if so how they are calculated? It could be that each kernel could define its own relative tolerance definition -- what do folks think about that?

MrBurmark: For reference tight is 1e-12, but that number is arbitrary and I don't really know if its a good number. All of the kernels using that tolerance have checksums that are identical on all of the platforms I've tested them on.

MrBurmark: I don't know if we want the docs to be where we define these values in case we decide to change them in the future.

docs/sphinx/dev_guide/kernel_class_impl.rst
MrBurmark: I took a pass at removing rep count dependencies at one point, that might be a good thing to look into again though.

PR #615, opened 2026-01-07
Filter with Variant Groups

MrBurmark:
# Summary

Add filtering by variant groups like `Base`, `Lambda` and `RAJA` as well as `Seq`, `OpenMP`, `Cuda`, and `Hip` to the variants input filtering options.

I also simplified some of the coding to use range based for loops.

Note that "Lambda" will filter variants like `Lambda_Seq` and `Kokkos_Lambda` which may not be desirable. Do people think that `Lambda` should filter `Kokkos_Lambda` or not?

- This PR is a feature
- It does the following:
  - Adds filtering on variant groups at the request of myself and others

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,10 +32,10 @@ terminology used in the Suite implementation and discussion includes:
     implementation approach and programming model, such as *baseline OpenMP*, 
     *RAJA OpenMP*, etc.
   * **Set** is a subset of variants in the Suite that exercise the same
-    implementation approach, such as Base with
-    Base_Seq, Base_OpenMP, and more using the *baseline* implementation
-    approach, or the same programming model, such as Seq with Base_Seq, Lambda_Seq, and RAJA_Seq using the
-    *Sequential* programming model.
+    implementation approach, such as Base with Base_Seq, Base_OpenMP, and more
+    using the *baseline* implementation approach, or the same programming model,
+    such as Seq with Base_Seq, Lambda_Seq, and RAJA_Seq using the *Sequential*
+    programming model.
   * **Tuning** refers to execution of kernels with a particular execution
     parameterization, such as GPU thread-block size 128, 256, etc.
 


docs/sphinx/dev_guide/structure.rst
```diff
@@ -12,9 +12,10 @@
 RAJA Performance Suite Structure: Adding Kernels, Variants, and Tunings 
 ************************************************************************
 
-This section describes how to add kernels, variants, groups, and tunings to the
-Suite. The discussion aims to make clear the organization of the code and 
-how it works, which is useful to understand when making a contribution.
+This section describes how to add kernels, kernel groups, variants, variant
```
rhornung67, line 15: I don't see how changing from "groups" to "kernel groups" make things clearer. It seems less clear to me.  Can you explain your rationale for the change?

docs/sphinx/dev_guide/structure.rst
```diff
@@ -202,6 +205,22 @@ adding a kernel ID and name, which is described in
...
+
+If a variant is added as part of a new group or new groups of variants in the
+Suite, new value(s) must be added to the ``VariantGroupID`` enum in the
+``RAJAPerfSuite.hpp`` header file and an associated variant group string name(s)
+must be added to the ``VariantGroupNames`` string array in the ``RAJAPerfSuite.cpp``
+source file. Adding a new variant group ID and name is essentially the same as
+adding a kernel group ID and name, which is described in
+:ref:`structure_addkernel_group-label`.
+
+.. note:: Enumeration values and string array entries for variant groups must be
```
rhornung67: I don't understand what is meant by "new groups of variants".

docs/sphinx/dev_guide/structure.rst
```diff
@@ -172,6 +173,8 @@ described above.
 Adding a variant to the Suite involves four main steps:
 
 #. Add a unique variant ID and a unique variant name to the Suite.
+#. If the variant is part of any new variant groups, add unique variant group
```
rhornung67: what is a "variant group"?

docs/sphinx/dev_guide/structure.rst
MrBurmark: This is to help differentiate groups of kernels from groups of variants.

docs/sphinx/dev_guide/structure.rst
MrBurmark: Should we add an example like the Base variant group contains variants like Base_Seq, Base_OpenMP, etc. and the Seq variant group contains variants like Base_Seq, Lambda_Seq, and RAJA_Seq?

rhornung67: Ah. I see what you are getting at now. To avoid overusing the term group, should we change "kernel group" to something like "kernel set" (club? gang? batch? troop?....) and similarly for "variant group".

MrBurmark: Its easy to change. I'm not sure that any of the other words is any less oversubscribed.

MrBurmark: We could use something like variant sets or subsets to keep the terms a bit further apart and avoid having to rename groups to kernel groups.

rhornung67: I like that better. I just did a CPU-only build and there is an unused variable warning in Executor.cpp. Should I push a change for that or will it disrupt what you are doing?

rhornung67: Also, adding a brief description about groups/sets like your comment above to the user docs would be good.

MrBurmark: @rhornung67 I changed the name to variant sets and added some more documentation.

docs/index.rst
```diff
@@ -31,6 +31,11 @@ terminology used in the Suite implementation and discussion includes:
   * **Variant** refers to implementations of Suite kernels that share the same 
     implementation approach and programming model, such as *baseline OpenMP*, 
     *RAJA OpenMP*, etc.
+  * **Set** is a subset of variants in the Suite that exercise the same
+    implementation approach or programming model, such as Base with the
```
rhornung67:
```suggestion
    implementation approach, such as Base with
```

docs/index.rst
```diff
@@ -31,6 +31,11 @@ terminology used in the Suite implementation and discussion includes:
...
+    Base_Seq, Base_OpenMP, and more using the *baseline* implementation
+    approach, and Seq with Base_Seq, Lambda_Seq, and RAJA_Seq using the
```
rhornung67:
```suggestion
    approach, or the same programming model, such as Seq with Base_Seq, Lambda_Seq, and RAJA_Seq using the
```

rhornung67 (changes requested):
One rewording suggestion. Hopefully, it makes it more clear.

PR #623, opened 2026-01-15
Add Memory Size Option

MrBurmark:
# Summary

Uses #622 to add a `--memory` option that chooses the first problem size that touches at least as much memory as requested.
This also adds a `--min-size` option to set a minimum size.
This also allows the `--sizefact` option to provide a multiplier on problem sizes with any other problem size options.

This makes it easy for example to benchmark memory bound kernels at a problem size where they are using main memory instead of the cache by asking for the amount of memory that is 4 times the size of the last level cache and a large enough problem size to use the full width of the hardware. If you want to turn that into a throughput study then `--sizefact` can be used to adjust the problem size from there.

- This PR is a feature
- It does the following:
  - Adds a memory based problem sizing option at the request of me

### DIFF
--- a/src/apps/CONVECTION3DPA.hpp
+++ b/src/apps/CONVECTION3DPA.hpp
@@ -8,7 +8,8 @@
 //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
 
 ///
-/// Action of 3D diffusion matrix via partial assembly
+/// Element-wise action of a 3D finite element volume convection operator
+/// via partial assembly and sum factorization
 ///
 /// Based on MFEM's/CEED algorithms.
 /// Reference implementation - MFEM-v4.9
@@ -17,8 +18,8 @@
 ///
 /// for(Index_type e = 0; e < NE; ++e) {
 ///
-///   constexpr Index_type max_D1D = CPA_D1D;
-///   constexpr Index_type max_Q1D = CPA_Q1D;
+///   constexpr Index_type max_D1D = conv::D1D;
+///   constexpr Index_type max_Q1D = conv::Q1D;
 ///   constexpr Index_type max_DQ = (max_Q1D > max_D1D) ? max_Q1D : max_D1D;
 ///   MFEM_SHARED Real_type sm0[max_DQ*max_DQ*max_DQ];
 ///   MFEM_SHARED Real_type sm1[max_DQ*max_DQ*max_DQ];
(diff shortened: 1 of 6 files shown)


src/common/RunParams.cpp
```diff
@@ -1474,16 +1503,23 @@ void RunParams::printHelpMessage(std::ostream& str) const
...
       << "\t\t --sizefact 2.0 (run each kernel with size twice its default size)\n\n";
 
   str << "\t --size <int> [no default]\n"
       << "\t      (kernel size to run for all kernels)\n"
-      << "\t      May not be set if --sizefact is set.\n";
+      << "\t      May not be set if --sizefact or --memory is set.\n";
   str << "\t\t Example...\n"
       << "\t\t --size 1000000 (runs each kernel with size ~1,000,000)\n\n";
 
+  str << "\t --memory <int> [no default]\n"
```
rhornung67, line 1520: We need to indicate unit for memory value (e.g., bytes).

MrBurmark: I added bytes higher up in the description. How is that?

rhornung67: Looks OK. I'm wondering if `memory touched to run with for all kernels` is clear or needs to be more descriptive. This only includes allocated array memory and is the aggregate summed over all arrays used in a kernel, right? Not sure how to improve it.

MrBurmark: It isn't the amount of memory allocated its the amount of memory "touched" in the kernel so it may be lower than the allocated amount of memory if it is not all touched (this is an issue for the comm kernels which don't touch all of their memory).

rhornung67: Got it. I was trying to think of ways folks could get confused about what the option means.

MrBurmark: I'll try to make this more clear.

MrBurmark: @rhornung67 I updated the help section again.

src/common/RunParams.cpp
```diff
@@ -1472,18 +1511,36 @@ void RunParams::printHelpMessage(std::ostream& str) const
...
-      << "\t      (kernel size to run for all kernels)\n"
-      << "\t      May not be set if --sizefact is set.\n";
+      << "\t      (problem size to run for all kernels)\n"
+      << "\t      May not be set if --memory is set.\n";
   str << "\t\t Example...\n"
       << "\t\t --size 1000000 (runs each kernel with size ~1,000,000)\n\n";
 
+  str << "\t --memory <int> [no default]\n"
+      << "\t      (each kernel calculates its problem size so that it touches\n"
+      << "\t       this many bytes of memory per rep)\n"
```
rhornung67:
```suggestion
      << "\t       the specified # of bytes of memory per rep)\n"
```

src/common/RunParams.hpp
```diff
@@ -361,8 +375,10 @@ class RunParams {
   double rep_fact;       /*!< pct of default kernel reps to run */
 
   SizeMeaning size_meaning; /*!< meaning of size value */
-  double size;           /*!< kernel size to run (input option) */
   double size_factor;    /*!< default kernel size multipier (input option) */
+  double size;           /*!< kernel size to run (input option) */
+  double memory;           /*!< memory size to run (input option) */
```
artv3, line 380: If `Factor` got removed above is `size_factor` still needed?

MrBurmark: It is now always applied. This doesn't break existing functionality and gives an easy way to scale the problem size. For example you can use `--memory` as a starting point for a throughput study and then scale that using `--sizefact` to larger or smaller problem sizes.

PR #625, opened 2026-01-16
FEM kernel description improvements

artv3:
Improvement to the kernel descriptions


rhornung67 (approved):
Thanks @artv3 I will update the kernel descriptions in the benchmarking doc

PR #628, opened 2026-01-20
FEM Kernel benchmarking

artv3:
# Summary

Integration of AMD's performance improvement for RAJA::launch and scripts for comparing performance through compilers and throughput.

### DIFF
--- a/src/apps/MASSVEC3DPA-Cuda.cpp
+++ b/src/apps/MASSVEC3DPA-Cuda.cpp
@@ -1,5 +1,5 @@
 //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
-// Copyright (c) Lawrence Livermore National Security, LLC and other 
+// Copyright (c) Lawrence Livermore National Security, LLC and other
 // RAJA Project Developers. See top-level LICENSE and COPYRIGHT
 // files for dates and other details. No copyright assignment is required
 // to contribute to RAJA Performance Suite.
@@ -91,7 +91,7 @@ __launch_bounds__(block_size) __global__ void MassVec3DPADirect(const Real_ptr B
   MASSVEC3DPA_0_GPU;
 
   MASSVEC3DPA_GPU_KERNEL_BODY(GPU_SHARED_DIRECT_2D, GPU_SHARED_DIRECT_3D);
-  
+
 }
 
 template <size_t block_size>
@@ -106,19 +106,11 @@ __launch_bounds__(block_size) __global__ void MassVec3DPALoop(const Real_ptr B,
   MASSVEC3DPA_0_GPU;
 
   MASSVEC3DPA_GPU_KERNEL_BODY(GPU_SHARED_LOOP_2D, GPU_SHARED_LOOP_3D);
-  
+
 }
 
 #define MASSVEC3DPA_CUDA_RAJA_LAUNCH                                           \
   {                                                                            \
-                                                                               \
-    constexpr bool async = true;                                               \
-                                                                               \
-    using launch_policy = RAJA::LaunchPolicy<                                  \
-    RAJA::cuda_launch_t<async, mvpa::Q1D * mvpa::Q1D * mvpa::Q1D>>;            \
-                                                                               \
-    using outer_x = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;               \
-                                                                               \
     /* clang-format off */                                                     \
     RAJA::launch<launch_policy>(                                               \
       res,                                                                     \
@@ -321,17 +313,25 @@ void MASSVEC3DPA::runCudaVariantImpl(VariantID vid)
 
     if constexpr (tune_idx == 0) {
 
+      constexpr bool async = true;
+
+      using launch_policy = RAJA::LaunchPolicy<
+      RAJA::cuda_launch_t<async, mvpa::Q1D * mvpa::Q1D * mvpa::Q1D>>;
+
+      using outer_x = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;
+
       using inner_x = RAJA::LoopPolicy<RAJA::cuda_thread_x_loop>;
 
       using inner_y = RAJA::LoopPolicy<RAJA::cuda_thread_y_loop>;
 
       using inner_z = RAJA::LoopPolicy<RAJA::cuda_thread_z_loop>;
 
+      using launch_context = RAJA::LaunchContext;
+
       startTimer();
       // Loop counter increment uses macro to quiet C++20 compiler warning
       for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
 
-        using launch_context = RAJA::LaunchContext;
         MASSVEC3DPA_CUDA_RAJA_LAUNCH;
 
       } // loop over kernel reps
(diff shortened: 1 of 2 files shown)


src/apps/MASS3DEA-Cuda.cpp
```diff
@@ -186,7 +225,35 @@ void MASS3DEA::runCudaVariantImpl(VariantID vid) {
...
+        if (vid == Base_CUDA) {
+          addVariantTuning<&MASS3DEA::runCudaVariantImpl<block_size, 0>>(
+              vid, "compile_block_inc_" + std::to_string(block_size));
+        }
+
+        if (vid == RAJA_CUDA) {
+          addVariantTuning<&MASS3DEA::runCudaVariantImpl<block_size, 0>>(
+              vid, "cache_loop_" + std::to_string(block_size));
+          addVariantTuning<&MASS3DEA::runCudaVariantImpl<block_size, 1>>(
+              vid, "compile_block_inc_" + std::to_string(block_size));
```
MrBurmark: Its nice if the corresponding tunings have the same tuning_idx.

src/apps/MASS3DEA-Hip.cpp
```diff
@@ -28,38 +28,120 @@ __global__ void Mass3DEA(const Real_ptr B, const Real_ptr D, Real_ptr M) {
...
 }
 
-template < size_t block_size >
-void MASS3DEA::runHipVariantImpl(VariantID vid) {
+template <typename inner_x, typename inner_y, typename inner_z, typename CONTEXT,
+          typename RESOURCE>
+void MASS3DEA::runRAJAImpl(RESOURCE &res)
+{
+
+  MASS3DEA_DATA_SETUP;
```
MrBurmark: This part in particular is something that I would like to avoid doing in each repetition. What about putting the whole loop body into a macro to avoid duplicating the implementation?

src/apps/FEM_MACROS.hpp
```diff
@@ -97,6 +101,24 @@
...
       for (Index_type tx = threadIdx.x; tx < Nx; tx += runtime_blocks_size)
 
+#define GPU_SHARED_2D_SWITCH(USE_DIRECT, tx, ty, Nx, Ny, BODY)                 \
+  do {                                                                         \
+    if constexpr (USE_DIRECT) {                                                \
+      GPU_SHARED_DIRECT_2D(tx, ty, Nx, Ny) { BODY; }                           \
+    } else {                                                                   \
+      GPU_SHARED_LOOP_2D(tx, ty, Nx, Ny) { BODY; }                             \
+    }                                                                          \
+  } while (false)
```
MrBurmark: I was thinking of doing this a different way. I worry that the this is something that could potentially change the compiler's decision making.

src/apps/MASSVEC3DPA-Hip.cpp
```diff
@@ -20,399 +20,210 @@
...
-{
-
-  const Index_type e = blockIdx.x;
-
-  MASSVEC3DPA_0_GPU;
-
-  GPU_SHARED_DIRECT_2D(q, d, mvpa::Q1D, mvpa::D1D) {
-    MASSVEC3DPA_1;
-  }
+  GPU_SHARED_2D_SWITCH(use_direct, q, d, mvpa::Q1D, mvpa::D1D, MASSVEC3DPA_1);
```
MrBurmark: I was thinking of doing something similar to what you did for the RAJA case for the base case. Including passing in some macros as arguments to the macro so you could switch between GPU_SHARED_LOOP_3D and GPU_SHARED_LOOP_3D_INC. You could probably have a single macro in the header for all of the Base variants and another for all of the RAJA variants, each with enough arguments to handle all of the differences.

artv3: thanks -- I think I got it with the latest push

src/apps/MASS3DEA-Cuda.cpp
```diff
@@ -28,38 +28,109 @@ __global__ void Mass3DEA(const Real_ptr B, const Real_ptr D, Real_ptr M) {
...
       }
     }
   }
 
 }
 
-template < size_t block_size >
-void MASS3DEA::runCudaVariantImpl(VariantID vid) {
+#define MASS3DEA_CUDA_RAJA_LAUNCH(RES, INNER_X, INNER_Y, INNER_Z, CONTEXT_T)   \
+  do {                                                                        \
```
MrBurmark: There is no need for the do while in RAJAPerf.

MrBurmark: Also in this case you don't really need to make arguments for the policies and resource since they are always the same.

src/apps/MASSVEC3DPA-Cuda.cpp
```diff
@@ -20,238 +20,85 @@
...
 
   MASSVEC3DPA_0_GPU;
 
-  GPU_SHARED_LOOP_2D(q, d, mvpa::Q1D, mvpa::D1D) {
-    MASSVEC3DPA_1;
+  if constexpr (use_direct) {
+    MASSVEC3DPA_GPU_KERNEL_BODY(GPU_SHARED_DIRECT_2D, GPU_SHARED_DIRECT_3D);
+  } else {
+    MASSVEC3DPA_GPU_KERNEL_BODY(GPU_SHARED_LOOP_2D, GPU_SHARED_LOOP_3D);
   }
```
MrBurmark: Please split this into two different global functions.

src/apps/MASSVEC3DPA-Cuda.cpp
MrBurmark: Use macro instead of function call.

src/apps/MASSVEC3DPA-Hip.cpp
```diff
@@ -580,29 +410,23 @@ void MASSVEC3DPA::defineHipVariantTunings()
         if (vid == Base_HIP) {
 
           addVariantTuning<&MASSVEC3DPA::runHipVariantImpl<block_size, 0>>(
-              vid, "BLOCKDIM_LOOP_INC_"+std::to_string(block_size));
+              vid, "blockdim_" + std::to_string(block_size));
```
MrBurmark: I'm still thinking about what to call these loops. How about something like `block_stride_loop`, then you could prepend things like `runtime_`, `cached_`, or `compiletime_` to differentiate how you're getting the block stride.

src/apps/MASSVEC3DPA-Cuda.cpp
```diff
@@ -20,400 +20,255 @@
...
-    MASSVEC3DPA_1;
+#define MASSVEC3DPA_CUDA_RAJA_LAUNCH                                           \
+  {                                                                            \
+                                                                               \
+    constexpr bool async = true;                                               \
+                                                                               \
+    using launch_policy = RAJA::LaunchPolicy<                                  \
+    RAJA::cuda_launch_t<async, mvpa::Q1D * mvpa::Q1D * mvpa::Q1D>>;            \
+                                                                               \
+    using outer_x = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;               \
```
MrBurmark: Can you move these outside of the macro? That's more consistent with the other FEM kernels.

src/apps/MASSVEC3DPA-Cuda.cpp
```diff
@@ -512,43 +331,51 @@ void MASSVEC3DPA::runCudaVariantImpl(VariantID vid)
       // Loop counter increment uses macro to quiet C++20 compiler warning
       for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
 
-        runRAJAImpl<inner_x, inner_y, inner_z>(res);
+        using launch_context = RAJA::LaunchContext;
```
MrBurmark: Can you move this outside of the irep loop for consistency as well?

MrBurmark (approved):

PR #634, opened 2026-01-23
Expand the Memory options

MrBurmark:
# Summary

Add --memory-touched and --memory-allocated options so you can get problem sizes for different memory counts.

This adds a BytesAllocated/Rep attribute/count.

- This PR is a feature
  - Adds more memory options at the request of myself

### DIFF
--- a/src/common/KernelBase.cpp
+++ b/src/common/KernelBase.cpp
@@ -95,7 +95,7 @@ KernelBase::KernelBase(KernelID kid, const RunParams& params)
                                          CALI_ATTR_ASVALUE |
                                          CALI_ATTR_AGGREGATABLE |
                                          CALI_ATTR_SKIP_EVENTS);
-  Bytes_Rep_attr = cali_create_attribute("BytesMoved/Rep", CALI_TYPE_INT,
+  Bytes_Moved_Rep_attr = cali_create_attribute("BytesMoved/Rep", CALI_TYPE_INT,
                                          CALI_ATTR_ASVALUE |
                                          CALI_ATTR_AGGREGATABLE |
                                          CALI_ATTR_SKIP_EVENTS);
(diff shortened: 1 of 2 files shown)


src/common/RunParams.cpp
```diff
@@ -1529,17 +1547,36 @@ void RunParams::printHelpMessage(std::ostream& str) const
...
   str << "\t --memory <int> [no default]\n"
-      << "\t      (# of bytes of memory each kernel will touch per rep; note that each kernel calculates a problem size so that it will\n"
-      << "\t        touch the given # of bytes of memory per rep)\n"
+      << "\t      (# of bytes of memory each kernel will \"use\" per rep; note that\n"
+      << "\t        each kernel calculates a problem size so that its\n"
+      << "\t        Bytes/rep will match the given # of bytes of memory per rep)\n"
+      << "\t      (kernels with fixed memory usage treat this the same as --size)\n"
+      << "\t      May not be set if --size or other --memory* option is set.\n";
+  str << "\t\t Example...\n"
+      << "\t\t --memory 1000000 (runs each kernel such that it has ~1,000,000 bytes per rep)\n\n";
```
rhornung67: So what does this mean now, bytes touched or bytes allocated?

I think `each kernel such that it has...` needs to be more specific. I'm not sure what it specifically means when a kernel has N bytes

MrBurmark: Neither, its trying to match the Bytes/Rep as output in the kernel details file. Which is different from either of touched or allocated each of which have their own arguments. We started with "Bytes/Rep", which is the number I use to figure out the effective bandwidth, then added more memory counts, see its entry under this heading of the documentation https://rajaperf.readthedocs.io/en/develop/sphinx/user_guide/output.html#kernel-details-output.

MrBurmark: It would be good to have a more specific name for "Bytes/Rep" instead of the generic name that it has now. Something like BytesTransferredToOrFromRAM/Rep would be descriptive but long maybe something like BytesTransferred/Rep or BytesMoved/Rep?

rhornung67: I like BytesMoved/Rep. It's not too long and is easy to understand.

src/common/KernelBase.cpp
```diff
@@ -90,7 +91,11 @@ KernelBase::KernelBase(KernelID kid, const RunParams& params)
                                            CALI_ATTR_ASVALUE |
                                            CALI_ATTR_AGGREGATABLE |
                                            CALI_ATTR_SKIP_EVENTS);
-  Bytes_Rep_attr = cali_create_attribute("Bytes/Rep", CALI_TYPE_INT,
+  Bytes_Allocated_Rep_attr = cali_create_attribute("BytesAllocated/Rep", CALI_TYPE_INT,
+                                         CALI_ATTR_ASVALUE |
+                                         CALI_ATTR_AGGREGATABLE |
+                                         CALI_ATTR_SKIP_EVENTS);
+  Bytes_Rep_attr = cali_create_attribute("BytesMoved/Rep", CALI_TYPE_INT,
```
MrBurmark: @michaelmckinsey1 I'm changing the name of `Bytes/Rep` to `BytesMoved/Rep` and have added more numbers like `BytesAllocated/Rep`.

PR #637, opened 2026-01-27
Rework build scripts to take optional cmake version

rhornung67:
# Summary

- This PR modifies build scripts to accept an optional cmake version argument
- If no cmake version is given, a default based on the RAJA cmake min version is set
- This helps with the fact that some of these scripts can be used on multiple platforms that may have different cmake version defaults or versions that don't exist on others

**We can improve these later. We want working scripts now!**

### DIFF
--- a/scripts/lc-builds/toss4_clang_caliper.sh
+++ b/scripts/lc-builds/toss4_clang_caliper.sh
@@ -9,6 +9,9 @@
 # SPDX-License-Identifier: (BSD-3-Clause)
 ###############################################################################
 
+# Default CMake version if not provided
+DEFAULT_CMAKE_VER=3.25.2
+
 if [[ $# -lt 3 ]]; then
   echo
   echo "You must pass 3 or more arguments to the script (in this order): "
--- a/scripts/lc-builds/toss4_cray-mpich_amdclang_asan.sh
+++ b/scripts/lc-builds/toss4_cray-mpich_amdclang_asan.sh
@@ -15,7 +15,7 @@ DEFAULT_CMAKE_VER=3.24.2
 if [[ $# -lt 3 ]]; then
   echo
   echo "You must pass 3 or more arguments to the script (in this order): "
-  echo "   1) cray-mpich compiler version number"
+  echo "   1) cray-mpich version number"
   echo "   2) HIP compiler version number"
   echo "   3) HIP compute architecture"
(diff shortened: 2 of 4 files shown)


scripts/lc-builds/toss4_clang_caliper.sh
```diff
@@ -11,34 +11,45 @@
...
 ADIAK_DIR=$3
-shift 3
+
+# Detect optional fourth positional argument as a CMake version if it looks like N.M or N.M.P
+# Otherwise, treat it as a normal CMake argument.
+if [ -n "$4" ] && [[ "$4" =~ ^[0-9]+(\.[0-9]+)*$ ]]; then
+  CMAKE_VER=$4
+  shift 4
+else
+  CMAKE_VER=$DEFAULT_CMAKE_VER
```
MrBurmark, line 38: Does this have a DEFAULT_CMAKE_VER?

rhornung67: Good catch. Fixed.

scripts/lc-builds/toss4_clang_caliper.sh
```diff
@@ -9,36 +9,50 @@
...
+  echo "You must pass 3 or more arguments to the script (in this order): "
   echo "   1) compiler version number"
   echo "   2) path to caliper cmake directory"
   echo "   3) path to adiak cmake directory"
+  echo "   4) optional CMake version to load."
   echo
   echo "For example: "
-  echo "    toss4_clang_caliper.sh 14.0.6 /usr/workspace/wsb/asde/caliper-quartz/share/cmake/caliper /usr/workspace/wsb/asde/caliper-quartz/lib/cmake/adiak"
-  exit
+  echo "    toss4_clang-mpi_caliper.sh 14.0.6 /usr/workspace/wsb/asde/caliper-quartz/share/cmake/caliper /usr/workspace/wsb/asde/caliper-quartz/lib/cmake/adiak  [3.27.4]"
```
smithsg84, line 24: The script is named  'toss4_clang_caliper.sh' no '-mpi'.  Perhaps use $(basename "$0") and eliminate the need to manually keep consistent?

```
SCRIPT_NAME="$(basename "$0")"

...

echo "    ${SCRIPT_NAME} 14.0.6 ...."

```

PR #638, opened 2026-01-29
Performance dashboard in CI

tdrwenski:
# Summary
Uses dedicated RADIUSS shared CI performance pipeline in to run performance benchmarks ~~regression tests~~. This runs on GitLab on LC machines (tuo and dane currently) ~~on every PR~~, collects timings using Caliper, puts average timing per rep into JSON format using a python script, and reports these to GitHub. GitHub receives the results as a workflow dispatch and uses the [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) to store ~~and compare to previous results~~. The results are only stored when run on `develop` ~~and other branches compare to the most recent value to give a pass/fail status on PRs, with a test threshold (currently 120%)~~. The results from `develop` will be viewable on this dashboard stored on github pages: https://software.llnl.gov/RAJAPerf/dev/bench/.

Edit: For now remove GitHub workflow pass/fail status on PRs, and only use dashboard functionality.

Updated version of @adrienbernede's https://github.com/llnl/RAJAPerf/pull/529, using GitLab components for CI.

TODO:

- [x]  in github workflow: `# if: github.ref == 'refs/heads/develop' # TODO uncomment before merging`
- [x] use radiuss-shared-ci release instead of branch in performance component once environment variable issue is fixed


scripts/gitlab/build_and_test.sh
```diff
@@ -290,7 +296,30 @@ then
...
+# Performance tests
+if [[ "${option}" != "--build-only" && "${perf_tests}" == "true" ]]
+then
+
+    cd ${project_dir}
+    mkdir -p ${perf_artifact_dir}
+
+    timed_message "Performance tests for RAJA Perf Suite"
+
+    raja_perf_command="${build_dir}/bin/raja-perf.exe --kernels ${perf_kernels} --exclude-kernels ${perf_exclude_kernels} --variants ${perf_variants} --outdir ${perf_artifact_dir} ${perf_run_opts}"
```
MrBurmark:
```suggestion
    raja_perf_command="${build_dir}/bin/raja-perf.exe --warmup-perfrun-same --kernels ${perf_kernels} --exclude-kernels ${perf_exclude_kernels} --variants ${perf_variants} --outdir ${perf_artifact_dir} ${perf_run_opts}"
```

.gitlab/jobs/performances.yml
```diff
@@ -0,0 +1,95 @@
...
+    PERF_VARIANTS: "RAJA_Seq"
+    # Exclude unstable tests (~20% variability):
+    PERF_EXCLUDE_KERNELS: >
+      ARRAY_OF_PTRS
+      COPY8
+      DAXPY
+      DAXPY_ATOMIC
+      REDUCE_STRUCT
+    PERF_RUN_OPTS: "--repfact 5.0"
+    PERF_MPI_EXEC: "srun --overlap -n 16"
```
MrBurmark: Why does this test set PERF_MPI_EXEC?

.gitlab/jobs/performances.yml
```diff
@@ -0,0 +1,95 @@
...
+      EMPTY
+      IF_QUAD
+      INDEXLIST_3LOOP
+      MAT_MAT_SHARED
+      MULTI_REDUCE
+      PI_REDUCE
+      REDUCE3_INT
+      REDUCE_STRUCT
+      TRAP_INT
+  extends: .perf_on_tuolumne
```
MrBurmark, line 43: Should we increase the problem size for HIP?
```suggestion
    PERF_RUN_OPTS: "--sizefact 5.0"
  extends: .perf_on_tuolumne
```

.gitlab/jobs/performances.yml
tdrwenski: it runs MPI with:
```
    if [[ -n ${perf_mpi_exec} ]]
    then
        raja_perf_command="${perf_mpi_exec} ${raja_perf_command}"
    fi
```
in the build and test script. not sure if this is the preferred way for RajaPerf or there is another way?

MrBurmark: If we're only running one job on the code, should we utilize all the cores?
```suggestion
    PERF_MPI_EXEC: "srun -n 112"
```

MrBurmark: Ah, I see this is the MPI test so it uses srun for that purpose while the others do not?

MrBurmark: I'm used to always using srun to run on a compute node even if I'm running without MPI `srun -n1`.

scripts/gitlab/caliper_to_github_benchmark.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Calculate average time per measurement (accounting for both reps and passes)
+        total_measurements = reps * passes
+        avg_time_per_measurement_ms = total_time_ms / total_measurements if total_measurements > 0 else total_time_ms
+
+        # Gather leaf-level kernel paths like "RAJAPerf/Group/KernelName"
+        if isinstance(path, list) and len(path) >= 3:
+            kernel_name = path[-1]
+            benchmark = {
+                "name": f"{variant}_{kernel_name}",
```
MrBurmark, line 51: I assume that here variant includes both the variant and tuning, so for example there are multiple "variants" for the HIP reduction kernels.

.gitlab/jobs/performances.yml
tdrwenski: yes, this is the only job at the moment built with `+mpi`. It is possible to always add `srun/ flux run` instead of just for this job but I don't know if that is clearer.

I think the `--overlap` is needed because we are running srun inside of the exclusive srun allocation for the CI. So the CI job runs with something like `srun --exclusive --reservation=ci --time=15 --nodes=1 build_and_test.sh` and then within that script it executes `srun raja-perf.exe ...`

We can use all cores but I believe it takes longer as then MPI is tied up in communication. I am not sure what is the best number of cores for testing these kernels, will leave that to you to decide.

scripts/gitlab/caliper_to_github_benchmark.py
tdrwenski: I believe so. If you look [here](https://software.llnl.gov/RAJAPerf/dev/bench/) at the test names there is RAJA_Seq-default and RAJA_Seq-new for instance

.gitlab/jobs/performances.yml
tdrwenski: I also chose `OMP_NUM_THREADS: "16"` for the openmp tests so we can alter that depending on what you want to test

.gitlab/jobs/performances.yml
tdrwenski: I will test this again. Previously it seemed to make the tests more variable but maybe with `warmup-perfrun-same` it could be different now

tdrwenski: I retested and this seems to help! added it now in https://github.com/llnl/RAJAPerf/pull/638/commits/f03bf6051022620bfaf2a12989c7c5b7a5e92794

.gitlab/jobs/performances.yml
tdrwenski: I retested and this seems to work well-- added in https://github.com/llnl/RAJAPerf/pull/638/commits/3124954aa92d7a58e220b975e74531528070fa46. Let me know if you want the same for openMP

MrBurmark: Its worth a try.

MrBurmark: Also probably increase the problem size for the openmp ones maybe set the `--sizefact 5` for openmp as well.

MrBurmark: For performance with openmp we probably only want to run openmp if we increase the problem size.

tdrwenski: Updated the openmp problem size to sizefact 5, left the `OMP_NUM_THREADS: "16"`, let me know if you wanted something else.

.gitlab/jobs/performances.yml
```diff
@@ -0,0 +1,56 @@
...
+# SPDX-License-Identifier: (BSD-3-Clause)
+##############################################################################
+
+clang_14_0_6_openmp_caliper:
+  variables:
+    SPEC: "~shared +openmp +caliper %%clang@=14.0.6"
+    PERF_KERNELS: "Basic"
+    PERF_VARIANTS: "RAJA_Seq RAJA_OpenMP"
+    PERF_RUN_OPTS: "--repfact 5.0 --sizefact 5.0"
+    OMP_NUM_THREADS: "16"
```
MrBurmark, line 15: Do we want to use OMP_NUM_THREADS or OMP_MAX_THREADS?

tdrwenski: Do you mean OMP_THREAD_LIMIT? My thought is that OMP_NUM_THREADS may be more deterministic but I'm not sure, so whatever you prefer for the first version.

MrBurmark: You are right, I think OMP_THREAD_LIMIT is preferred as it actually tells omp the max instead of just saying how many to use but leaving the max to omp.

rhornung67:
@tdrwenski I'm working on getting a RAJA Perf release out this week. Please hold off on merging this until that is done. Thank you.

PR #642, opened 2026-02-05
Update RSCI and adapt to machine-status-override

adrienbernede:
This fixes the issue where a failure status due to machine outage is never overridden when the machine comes back and CI passed.


adrienbernede:
> The reason that the performance check is failing appears to be a missing file. GitLab complains that `.gitlab/jobs/performances.yml` does not exist

Indeed. But I was ignoring this since it was not preventing me from running my tests. It turns out this comes from a muddle when creating the branch. The performance pipeline is setup in #638.

PR #647, opened 2026-02-13
Add Missing CLI Params to Adiak Metadata

michaelmckinsey1:
# Summary

- This PR is a feature
  - Modifies `LTIMES.cpp` and `src/common/Executor.cpp`
  - Adds kernel-specific metadata at the request of @michaelmckinsey1 

- [x] ~Change the way zones is computed, such that the problem size you set on the command line will end up being the amount of unknowns (per-process). I did this so it is easier to compare against Kripke. I'm curious what @MrBurmark thinks; I do not know why the original expression is `(target_size + (m_num_d * m_num_g)/2) / (m_num_d * m_num_g)`. If there is a specific reason, please let me know. I think it would be nice if we can simplify it.~
- [x] Introduce metadata specific to the LTIMES kernel. ~LTIMES already has CLI-specific arguments  `--ltimes_num_m`, `--ltimes_num_g`, `--ltimes_num_d`. This might be better implemented as just adding those to the metadata, and with the new definition of Z, it is easy to compute from the metadata after the run~ Changed this in favor of adding the CLI arguments to the metadata, as we do for other CLI arguments.
   - Z can be computed from `problemsize/g/d`. Total amount of unknowns is psilen or `problemsize`
- [x] Change default values for LTimes to match an experimental configuration. 1M default problem size is fine.

### DIFF
--- a/src/common/RunParams.cpp
+++ b/src/common/RunParams.cpp
@@ -44,8 +44,8 @@ RunParams::RunParams(int argc, char** argv)
    data_alignment(RAJA::DATA_ALIGN),
    multi_reduce_num_bins(10),
    multi_reduce_bin_assignment_algorithm(BinAssignmentAlgorithm::RunsRandomSizes),
-   ltimes_num_d(12),
-   ltimes_num_g(220),
+   ltimes_num_d(6),
+   ltimes_num_g(32),
    ltimes_num_m(25),
    array_of_ptrs_array_size(ARRAY_OF_PTRS_MAX_ARRAY_SIZE),
    halo_width(1),
@@ -1594,13 +1594,13 @@ void RunParams::printHelpMessage(std::ostream& str) const
       << "\t\t --sizefact 0.5 (run each kernel with size half its calculated size)\n"
       << "\t\t --sizefact 2.0 (run each kernel with size twice its calculated size)\n\n";
 
-  str << "\t --ltimes_num_d <int> [default is 12]\n"
+  str << "\t --ltimes_num_d <int> [default is 6]\n"
       << "\t      (For LTIMES kernels only: num_d used in kernels)\n"
       << "\t      Must be greater than 0.\n";
   str << "\t\t Example...\n"
       << "\t\t --ltimes_num_d 32\n\n";
 
-  str << "\t --ltimes_num_g <int> [default is 220]\n"
+  str << "\t --ltimes_num_g <int> [default is 32]\n"
       << "\t      (For LTIMES kernels only: num_g used in kernels)\n"
       << "\t      Must be greater than 0.\n";
   str << "\t\t Example...\n"


src/apps/LTIMES.cpp
```diff
@@ -52,7 +52,7 @@ LTIMES::LTIMES(const RunParams& params)
 
 void LTIMES::setSize(Index_type target_size, Index_type target_reps)
 {
-  m_num_z = std::max((target_size + (m_num_d * m_num_g)/2) / (m_num_d * m_num_g), Index_type(1));
+  m_num_z = std::max((target_size/m_num_d/m_num_g), Index_type(1));
```
MrBurmark: This is done because `num_d` and `num_g` are fixed so the problem size must be a multiple of `num_d*num_g`. This expression allows the problem size to be some times greater and sometimes lesser than the target problem size but on average closer to the target problem size given on the command line instead of on average being lesser.

src/common/RunParams.cpp
```diff
@@ -44,8 +44,8 @@ RunParams::RunParams(int argc, char** argv)
    data_alignment(RAJA::DATA_ALIGN),
    multi_reduce_num_bins(10),
    multi_reduce_bin_assignment_algorithm(BinAssignmentAlgorithm::RunsRandomSizes),
-   ltimes_num_d(64),
-   ltimes_num_g(32),
+   ltimes_num_d(12),
+   ltimes_num_g(220),
    ltimes_num_m(25),
```
MrBurmark: I'm not sure why the default values were chosen as they were. We'll have to ask @rchen20 what reasonable defaults are.

rhornung67: I set the defaults a long time ago based on a conversation I had with Adam K. We should probably pick defaults based on a typical DT problem or maybe what Kripke uses for defaults.

michaelmckinsey1: https://github.com/llnl/Kripke/blob/01f6f85c02ceffcd2bc06e42cee997867dd142c5/src/Kripke/InputVariables.cpp#L18-L21 These are the defaults. 

The values I am using are based on a typical problem, to my knowledge.

If we wanted to use Kripke defaults - `m` in kripke is `legendre_order + 1)^2`, so `m=25` matches the kripke default `legendre_order=4`. 220 groups is the smallest out of a set `{220, 320, 360}` that we have been running (you will get very different performance running 32 or 64 groups), but to match kripke we would leave this at `32`. Where it gets tricky is the number of directions is decomposed in Kripke for LTimes, so if we actually want to match the Kripke parameter for directions - it would be `96 / num_dirsets / num_groupsets` =  `96 / 8 / 2` = `6` (I set `12` because I was using `groupsets=1`).

So if we want to match the defaults it would then be:
```
   ltimes_num_d(6),
   ltimes_num_g(32),
   ltimes_num_m(25),
```

Zones would almost always be different, since this will depend on problem size and number of processes in RAJAPerf. 

rajaperf - (1M + 19) / 192 = 5461 (per-process zones)
kripke - 16x16x16 = 4096 (total zones)

src/apps/LTIMES.cpp
michaelmckinsey1: I have reverted this change

src/common/RunParams.cpp
michaelmckinsey1: @rhornung67 @MrBurmark I have changed the defaults again, such that 1 rep of LTimes is comparable to 1 kernel run in Kripke, with default parameters.

MrBurmark: That seems reasonable, but I'll double check with @rchen20.

rchen20: > @rhornung67 @MrBurmark I have changed the defaults again, such that 1 rep of LTimes is comparable to 1 kernel run in Kripke, with default parameters.

12 for directions, and 25 for m seem good. Why are we setting groups to 220? Is this to match the number of CUs on the MI300A?

michaelmckinsey1: @rchen20 This is outdated, see https://github.com/llnl/RAJAPerf/pull/647/changes. I set d=6, because the default kripke decomposition is gset=2 and 96 total angles. groups=220 was for that reason, but I changed it back to match the kripke default value, so ignore that change.

But I noticed with gset=2, performance can vary up to 50% between kripke and rajaperf LTimes. For gset=1 (so d=12), they are within 5% difference. So we run with `gset=1` for performance comparisons. So I would actually think it would make more sense to set d=12 here, and gset=1 in kripke by default, but I'm guessing gset=1 is not a good default.

michaelmckinsey1: tldr; the way it is currently set in this PR matches the current kripke defaults.

rchen20: Reflecting on this some more, I would guess that `gset=2` was set as the default in Kripke because if a user tries different layouts (e.g. GDZ, ZGD, etc.), there would be non-zero sets of groups that can be iterated over, thus providing some level of parallelism in the group dimension regardless of where the group loop takes place. Even though there is a large performance difference, I think we should preserve `gset=2` in RAJA Perf, so we'd want `d=6`.

Does the performance of `gset=2, d=6` become more consistent if you increase the zone count? That's what I would expect to happen.

michaelmckinsey1: (shared some results on teams)

michaelmckinsey1: We decided to stick with `d=6`, with the caveat that decreasing gset by 2x is not equivalent to decreasing d by 2x. So for performance you are better off setting gset=1 in kripke and d=12 in rajaperf (if d=96 in kripke)

rhornung67:
@michaelmckinsey1 you are correct. i set up the branch protection rules that way. i will merge.

PR #653, opened 2026-03-05
Internal Mesh Generator for FEMSWEEP

rchen20:
# Summary

- This PR is a feature
- It does the following:
  - Adds mesh generator for the FEMSWEEP, which allows meshes to be generated from the PerfSuite command line. Users can specify:

1.  Polar angles (default 3)
2. Azimuthal angles (default 3)
3. Energy groups (default 16)
4. Mesh elements in x, y, and z directions (default TBD, but 5,5,5 may be a fast running simulation)

### DIFF
--- a/src/apps/FEMSWEEPMeshGen.cpp
+++ b/src/apps/FEMSWEEPMeshGen.cpp
@@ -17,8 +17,11 @@
 namespace rajaperf
 {
 
+namespace numbers
+{
 // Universal definition of pi
-inline constexpr Real_type RAJAPERF_PI = 3.1415926535897932385;
+inline constexpr Real_type pi = 3.1415926535897932385;
+}
 
 namespace apps
 {
@@ -34,7 +37,7 @@ static void GaussLegendre(Index_type n, std::vector<Real_type> &x, std::vector<R
    const Index_type m = (n + 1) / 2;
    for (Index_type i = 0; i < m; ++i)
    {
-      Real_type z = std::cos(RAJAPERF_PI * (i + 0.75) / (n + 0.5));
+      Real_type z = std::cos(rajaperf::numbers::pi * (i + 0.75) / (n + 0.5));
       Real_type z1 = 0.0;
       while (true)
       {
@@ -91,7 +94,7 @@ AngularQuadratureLite::AngularQuadratureLite(Index_type polar_order, Index_type 
       const Real_type theta = std::acos(x[static_cast<Size_type>(j)]);
       for (Index_type k = 0; k < azimuthalAngles; ++k)
       {
-         const Real_type gamma = (-RAJAPERF_PI / azimuthalAngles) + (2.0 * RAJAPERF_PI / azimuthalAngles) * (k + 1.0);
+         const Real_type gamma = (-rajaperf::numbers::pi / azimuthalAngles) + (2.0 * rajaperf::numbers::pi / azimuthalAngles) * (k + 1.0);
          const Real_type st = std::sin(theta);
          omega_.push_back({st * std::cos(gamma), st * std::sin(gamma), std::cos(theta)});
       }


src/apps/FEMSWEEP.cpp
```diff
@@ -218,66 +138,62 @@ void FEMSWEEP::setUp(VariantID vid, size_t RAJAPERF_UNUSED_ARG(tune_idx))
   allocAndInitDataRandValue (m_M0dat    , m_M0len     , vid);
   allocAndInitDataRandValue (m_Xdat     , m_Xlen      , vid);
 
-  // Read mesh connectivity data from file.
-  std::ifstream dataFile(m_mesh_file);
+  auto temp_arr = allocDataForInit(m_nhpaa_r, m_sweeper->m_nhyperplanes_all_angles.Size(), vid);
+  for ( int ii = 0; ii < m_sweeper->m_nhyperplanes_all_angles.Size(); ++ii )
+  {
+    m_nhpaa_r[ii] = m_sweeper->m_nhyperplanes_all_angles[ii];
+  }
```
MrBurmark:
```suggestion
  allocAndCopyHostData(m_nhpaa_r, m_sweeper->m_nhyperplanes_all_angles.data(), m_sweeper->m_nhyperplanes_all_angles.size(), vid);
```
You can probably do something like this to avoid the extra code here.

src/common/RunParams.cpp
```diff
@@ -1636,6 +1759,42 @@ void RunParams::printHelpMessage(std::ostream& str) const
...
+  str << "\t\t Example...\n"
+      << "\t\t --femsweep_angles_polar 3\n\n";
+
+  str << "\t --femsweep_angles_azim <int> [default is 3]\n"
+      << "\t      (For FEMSWEEP kernel only)\n"
+      << "\t      Must be greater than 0.\n";
+  str << "\t\t Example...\n"
+      << "\t\t --femsweep_angles_azim 3\n\n";
+
+  str << "\t --femsweep_groups <int> [default is 128]\n"
```
MrBurmark: Is this default correct?

src/apps/FEMSWEEPMeshGen.hpp
```diff
@@ -0,0 +1,216 @@
...
+  public:
+   MDIntArray() = default;
+   MDIntArray(int d0, int d1) : d0_(d0), d1_(d1), data_(static_cast<size_t>(d0 * d1)) {}
+   int Size() const { return static_cast<int>(data_.size()); }
+   const int *HostRead() const { return data_.data(); }
+   int &operator()(int i0, int i1) { return data_.at(static_cast<size_t>(i0 + d0_ * i1)); }
+   int operator()(int i0, int i1) const { return data_.at(static_cast<size_t>(i0 + d0_ * i1)); }
+
+  private:
+   int d0_ = 0, d1_ = 0;
```
MrBurmark: Does it make sense to use the RAJAPerf Index_Type for all of the dimensions? If not it probably makes more sense to cast to size_t before doing arithmetic instead of after to avoid potential overflows.

src/apps/FEMSWEEPMeshGen.cpp
```diff
@@ -0,0 +1,749 @@
...
+{
+namespace apps
+{
+
+// Derived from the standard Gauss–Legendre quadrature construction:
+// - The original MFEM-based `AngularQuadrature` used `mfem::IntRules.Get(SEGMENT, 2*polarAngles-1)`
+//   which returns an n-point Gauss–Legendre rule on [-1,1].
+// - Here we implement the same rule directly via Newton iterations on Legendre polynomials.
+// This is a common textbook algorithm (e.g. Numerical Recipes), not copied from MFEM code.
+static void GaussLegendre(int n, std::vector<double> *x, std::vector<double> *w)
```
MrBurmark: Why not take the vectors by reference?

src/apps/FEMSWEEPMeshGen.cpp
```diff
@@ -0,0 +1,749 @@
...
+// - Same formula for direction vectors:
+//     omega = (sin(theta)cos(gamma), sin(theta)sin(gamma), cos(theta))
+// We intentionally omit weights and spherical harmonics since the mesh generator only needs omega vectors.
+AngularQuadratureLite::AngularQuadratureLite(int polar_order, int azimuthal_order)
+{
+   const int polarAngles = 2 * polar_order;
+   const int azimuthalAngles = 4 * azimuthal_order; // 3D
+
+   std::vector<double> x, w;
+   GaussLegendre(polarAngles, &x, &w);
```
MrBurmark: Don't x and w need to be setup before this call?

src/apps/FEMSWEEP.cpp
rchen20: Thanks, that's much better than what I was doing. Changed.

src/common/RunParams.cpp
rchen20: Good catch, it's actually 16. Fixed.

src/apps/FEMSWEEPMeshGen.hpp
rchen20: Let me think about this a bit. I've changed the connectivity arrays to be `Int_ptr` to match all the index arithmetic happening in the sweep. Unfortunately, one of the indices can have negative values, so keeping things as `int` may be the way to go here.

src/apps/FEMSWEEPMeshGen.cpp
rchen20: Done.

src/apps/FEMSWEEPMeshGen.cpp
rchen20: They were being initialized in the `GaussLegendre()` function itself, but I've moved that initialization to the more logical place here.

src/apps/FEMSWEEP.cpp
```diff
@@ -27,63 +28,16 @@ namespace apps
 FEMSWEEP::FEMSWEEP(const RunParams& params)
   : KernelBase(rajaperf::Apps_FEMSWEEP, params)
 {
-  auto readOneLong = [=](std::ifstream & file, Index_type & val, std::string valname) {
+  m_na = 8 * params.getFemsweepPolar() * params.getFemsweepAzim();
```
rhornung67: Should the problem size and mesh setup be done in a `setSize()` function instead of the ctor? That is the implementation pattern that we have in all the other kernel classes. For example, https://github.com/llnl/RAJAPerf/blob/develop/src/apps/VOL3D.cpp#L48

src/common/RunParams.cpp
```diff
@@ -1636,6 +1759,42 @@ void RunParams::printHelpMessage(std::ostream& str) const
...
+  str << "\t\t Example...\n"
+      << "\t\t --femsweep_ny 5\n\n";
+
+  str << "\t --femsweep_nz <int> [default is 5]\n"
+      << "\t      (For FEMSWEEP kernel only)\n"
+      << "\t      Must be greater than 0.\n";
+  str << "\t\t Example...\n"
+      << "\t\t --femsweep_nz 5\n\n";
+
   str << "\t --femsweep-mesh-file, -fmf <string> [Default is no file, and the femsweep is not run. The femsweep can only be run by specifying this option with a valid path to a mesh file name.]\n"
```
rhornung67: The help section for the femsweep mesh file should be removed since that input option has been removed.

rhornung67 (changes requested):

src/apps/FEMSWEEP.cpp
rchen20: I think we'll need to keep it the way it is. In order to set those parameters (for the mesh generator), I need access to the `params` object. The `params` object is not passed into `setSize()` and it's a pure virtual function, so I'd need to change its definition in `KernelBase`. We can just keep it the way it is because `setSize()` is setting all the other array sizes that really matter to the simulation, while what's happening in the constructor is purely setting the mesh.

src/common/RunParams.cpp
rchen20: Thanks for catching this, it's been removed.

src/apps/FEMSWEEPMeshGen.hpp
MrBurmark: I would prefer using the rajaperf type aliases everywhere, for example `int` is `Int_type`.

MrBurmark: Note that `Int_type` is a 32-bit signed integer type and `Index_type` is a signed 64-bit integer type.

src/apps/FEMSWEEPMeshGen.cpp
```diff
@@ -0,0 +1,748 @@
...
+// to contribute to RAJA Performance Suite.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include "FEMSWEEPMeshGen.hpp"
+
+// Universal definition of pi
+#ifndef PI_
+#define PI_ 3.1415926535897932385
```
MrBurmark: Its best practice to prepend macros with the code name, e.g. `RAJAPERF_PI`

MrBurmark: Or even better make this a constexpr global variable instead.

src/apps/FEMSWEEPMeshGen.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+// A multi-dimensional integer array with 2D indexing and a contiguous "HostRead()" view,
+// matching what the original code used from MFEM's `mfem::MDArray<int,2>`.
+template <> class MDIntArray<2>
+{
+  public:
+   MDIntArray() = default;
+   MDIntArray(int d0, int d1) : d0_(d0), d1_(d1), data_(static_cast<size_t>(d0 * d1)) {}
+   int Size() const { return static_cast<int>(data_.size()); }
+   const int *HostRead() const { return data_.data(); }
```
MrBurmark: Can we just get away with a Data method and remove HostRead?

src/apps/FEMSWEEPMeshGen.cpp
rhornung67: Yes, please make it a constexpr variable inside the rajaperf namespace

src/apps/FEMSWEEPMeshGen.hpp
rhornung67: @rchen20 note that all the data type aliases are defined in this file: https://github.com/llnl/RAJAPerf/blob/develop/src/common/RPTypes.hpp#L60

src/common/RunParams.cpp
rhornung67: @rchen20 I still see it in this PR and in the code in your branch. Did you push the change?

rchen20: I just pushed another change. Previously, I had removed some mesh file reading code somewhere else.

src/apps/FEMSWEEP.cpp
rchen20: This has been refactored to create the mesh in `setSize()`.

src/apps/FEMSWEEPMeshGen.cpp
rchen20: Done.

src/apps/FEMSWEEPMeshGen.hpp
rchen20: Done.

src/apps/FEMSWEEPMeshGen.hpp
rchen20: Done.

src/apps/FEMSWEEP.cpp
```diff
@@ -108,8 +52,43 @@ FEMSWEEP::FEMSWEEP(const RunParams& params)
...
-void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type target_reps)
+void FEMSWEEP::setSize(Index_type target_size, Index_type target_reps)
 {
+  // Set basic mesh parameters
+
+  // Always use user or default angles and groups.
+  m_na = 8 * this->run_params.getFemsweepPolar() * this->run_params.getFemsweepAzim();
+  m_ng = this->run_params.getFemsweepGroups();
+
+  if (this->run_params.getSizeMeaning() == RunParams::SizeMeaning::Memory)
```
MrBurmark: The only thing that matters is target_size, the behavior should not differ based on the size meaning.

src/apps/FEMSWEEP.cpp
```diff
@@ -108,8 +52,43 @@ FEMSWEEP::FEMSWEEP(const RunParams& params)
...
+      m_ny = rounded_cube;
+      m_nz = rounded_cube;
+    }
+
+  }
+  else  // Using user or default parameters to set mesh size.
+  {
+    m_nx = this->run_params.getFemsweepX();
+    m_ny = this->run_params.getFemsweepY();
+    m_nz = this->run_params.getFemsweepZ();
```
MrBurmark, line 79: Is there something like `isFemsweepXSet` so we can choose to use the number or generate our own number?

src/apps/FEMSWEEPMeshGen.hpp
```diff
@@ -0,0 +1,215 @@
...
+// This header defines small array containers and a `MeshGenerator` that reimplements those pieces
+// using logic derived from MFEM and the previous MFEM-backed transport code:
+// - MFEM mesh topology / face numbering:
+//   `mfem_git_v4.7/mesh/mesh.cpp` (`MakeCartesian3D`, `GetElementToFaceTable`, `GetQuadOrientation`)
+//   and `mfem_git_v4.7/general/stable3d.cpp` (`STable3D::Push4` keying rule).
+// - MFEM Hilbert SFC ordering:
+//   `mfem_git_v4.7/mesh/ncmesh.cpp` (`NCMesh::GridSfcOrdering3D`, `HilbertSfc3D`).
+// - Transport code GPU setup structure:
+//   `transport_code/src/sweep/gpu/setup/{Connectivity,FaceTypes,Hyperplanes}.cc`.
+class IntArray
```
artv3, line 70: You can probably just derive from std::Vector? then you don't need to add the methods.

src/apps/FEMSWEEPMeshGen.cpp
```diff
@@ -0,0 +1,747 @@
...
+#include <algorithm>
+#include <cmath>
+#include <cstdint>
+#include <unordered_map>
+
+namespace rajaperf
+{
+
+// Universal definition of pi
+inline constexpr Real_type RAJAPERF_PI = 3.1415926535897932385;
```
artv3: Minor; I would be more in favor of using `M_PI` or once we move to c++20, `std::numbers::pi;`

src/apps/FEMSWEEPMeshGen.hpp
rhornung67: That's generally a bad idea since std::vector does not have a virtual destructor by design, which is an important feature of a C++ base class.

artv3: Ah, gotcha. I missed that. Thanks for pointing that out.

MrBurmark: But does it really matter if you're not going to use virtual inheritance?

MrBurmark: You can privately inherit, then add some explicit using statements to use some of the base class methods.

src/apps/FEMSWEEPMeshGen.cpp
MrBurmark: How about just `pi` since this is now in a namespace and not a macro. Or maybe imitate what the standard does and add a `numbers` namespace as well so its `rajaperf::numbers::pi`?

artv3: The later sounds good `rajaperf::numbers::pi` but I was just thinking we probably do not need to redefine established constants.

src/common/RunParams.cpp
```diff
@@ -47,6 +47,11 @@ RunParams::RunParams(int argc, char** argv)
    ltimes_num_d(6),
    ltimes_num_g(32),
    ltimes_num_m(25),
+   femsweep_angles_polar(3),
+   femsweep_angles_azim(3),
+   femsweep_groups(16),
+   femsweep_mesh_dims({5, 5, 5}),
+   is_femsweep_mesh_dims(false),
```
MrBurmark: How about `use_femsweep_mesh_dims` or `have_femsweep_mesh_dims`?

src/common/RunParams.cpp
```diff
@@ -147,6 +151,14 @@ void RunParams::print(std::ostream& str) const
   str << "\n ltimes_num_g = " << ltimes_num_g;
   str << "\n ltimes_num_m = " << ltimes_num_m;
 
+  str << "\n femsweep_angles_polar = " << femsweep_angles_polar;
+  str << "\n femsweep_angles_azim = " << femsweep_angles_azim;
+  str << "\n femsweep_groups = " << femsweep_groups;
+  str << "\n femsweep_mesh_dims = ";
+  for (size_t j = 0; j < femsweep_mesh_dims.size(); ++j) {
+    str << "\n\t" << femsweep_mesh_dims[j];
+  }
```
MrBurmark, line 161: Also output `is_femsweep_mesh_dims`.

src/apps/FEMSWEEP.cpp
```diff
@@ -108,8 +52,44 @@ FEMSWEEP::FEMSWEEP(const RunParams& params)
...
+      m_ny = 1;
+      m_nz = 1;
+    }
+    else
+    {
+      Index_type rounded_cube = std::round( std::cbrt(remainder) );
+      m_nx = rounded_cube;
+      m_ny = rounded_cube;
+      m_nz = rounded_cube;
+    }
```
MrBurmark: It would be fewer lines to do something like `std::max(1.0, std::round(...));`

src/apps/FEMSWEEP.cpp
```diff
@@ -152,14 +132,14 @@ void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type t
...
                       1*sizeof(Index_type) * NLF * m_ne + // elem_to_faces
-                      1*sizeof(Index_type) * 10800 + // F_g2l
-                      1*sizeof(Index_type) * 37800 + // idx1
-                      1*sizeof(Index_type) * 37800 );// idx2
-  setBytesWrittenPerRep( 0 );
+                      1*sizeof(Index_type) * (m_sharedinteriorfaces + m_boundaryfaces) + // F_g2l
+                      1*sizeof(Index_type) * m_sharedinteriorfaces * 4 + // idx1
+                      1*sizeof(Index_type) * m_sharedinteriorfaces * 4 );// idx2
+  setBytesWrittenPerRep( 1*sizeof(Real_type) * m_Xlen );  // Xdat
   setBytesModifyWrittenPerRep( 1*sizeof(Real_type) * m_Xlen ); // Xdat
```
MrBurmark: Anything counted in ModifyWritten shouldn't be counted in Read or Written.

src/common/RunParams.cpp
artv3: I agree with @MrBurmark  suggestions here

src/apps/FEMSWEEP.hpp
MrBurmark: Use the type aliases here.

src/apps/FEMSWEEP.cpp
```diff
@@ -137,147 +117,81 @@ void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type t
...
-                      1*sizeof(Index_type) * m_na * 43 + // phpaa_r,
-                      1*sizeof(Index_type) * m_na * m_ne + // order_r,
-                      1*sizeof(Index_type) * NLF * m_ne * m_na + // AngleElem2FaceType
-                      1*sizeof(Index_type) * NLF * m_ne + // elem_to_faces
-                      1*sizeof(Index_type) * 10800 + // F_g2l
-                      1*sizeof(Index_type) * 37800 + // idx1
-                      1*sizeof(Index_type) * 37800 );// idx2
-  setBytesWrittenPerRep( 0 );
-  setBytesModifyWrittenPerRep( 1*sizeof(Real_type) * m_Xlen ); // Xdat
+  setBytesReadPerRep( m_ng*2*sizeof(Real_type) * m_Blen + // Bdat
```
MrBurmark: This should only count once for each address read. In other words it assumes perfect caching within each loop.

src/apps/FEMSWEEP.cpp
```diff
@@ -137,147 +117,81 @@ void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type t
...
+                           1*sizeof(Index_type) * (m_sharedinteriorfaces + m_boundaryfaces) + // F_g2l
+                           1*sizeof(Index_type) * m_sharedinteriorfaces * 4 + // idx1
+                           1*sizeof(Index_type) * m_sharedinteriorfaces * 4 );// idx2
   // using total data size instead of writes and reads
   setBytesReadPerRep( 1*sizeof(Real_type) * m_Blen + // Bdat
                       1*sizeof(Real_type) * m_Alen + // Adat
                       1*sizeof(Real_type) * m_Flen + // Fdat
                       1*sizeof(Real_type) * m_Sglen + // Sgdat
                       1*sizeof(Real_type) * m_M0len + // M0dat
+                      (ND-1)*sizeof(Real_type) * m_na * m_ng * m_ne + // Xdat
```
MrBurmark: Are you reading and writing to Xdat?

src/apps/FEMSWEEP.cpp
```diff
@@ -152,132 +132,65 @@ void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type t
...
-                      1*sizeof(Index_type) * 10800 + // F_g2l
-                      1*sizeof(Index_type) * 37800 + // idx1
-                      1*sizeof(Index_type) * 37800 );// idx2
-  setBytesWrittenPerRep( 0 );
-  setBytesModifyWrittenPerRep( 1*sizeof(Real_type) * m_Xlen ); // Xdat
+                      1*sizeof(Index_type) * (m_sharedinteriorfaces + m_boundaryfaces) + // F_g2l
+                      1*sizeof(Index_type) * m_sharedinteriorfaces * 4 + // idx1
+                      1*sizeof(Index_type) * m_sharedinteriorfaces * 4 );// idx2
+  setBytesWrittenPerRep( 1*sizeof(Real_type) * (m_Xlen - 1.55 * m_na * m_ng * m_sharedinteriorfaces) ); // remainder of Psi after sharedinteriorfaces
+  setBytesModifyWrittenPerRep( 1.55*sizeof(Real_type) * m_na * m_ng * m_sharedinteriorfaces );  // a multiple of sharedinteriorfaces
```
MrBurmark: How about making the number of modifywritten locations a local variable so its more clear what 's going on. 
```suggestion
  Index_type amount_of_x_modify_written = 1.55 * m_na * m_ng * m_sharedinteriorfaces; // this is an estimate
  setBytesWrittenPerRep( 1*sizeof(Real_type) * (m_Xlen - amount_of_x_modify_written ) ); // Xdat
  setBytesModifyWrittenPerRep( sizeof(Real_type) * amount_of_x_modify_written ); // Xdat
```

src/apps/FEMSWEEP.hpp
```diff
@@ -119,74 +122,79 @@ constexpr int FDS = 4;  // number of DOFs per face
...
 #define FEMSWEEP_KERNEL \
-  const int a = ag / ng; \
-  const int g = ag % ng; \
-  const int nhp = nhpaa_r[a]; \
-  const int ohp = ohpaa_r[a]; \
-  int s_nehp_done = 0; \
-  double A[ND * ND], b[ND]; \
-  double Ffactor = fmax(sin(Adat[order_r[a*ne]*ND*ND + a*ne*ND*ND]) - 2.0, 0.0); \
-  for (int hp = 0; hp < nhp; ++hp) \
+  const Int_type a = ag / ng; \
```
MrBurmark: Do you want Int_type or Index_type, the arrays contain Index_types.

src/apps/FEMSWEEP.cpp
rchen20: Done.

src/apps/FEMSWEEP.hpp
rchen20: Done.

src/common/RunParams.cpp
rchen20: Done.

src/apps/FEMSWEEP.cpp
rchen20: Done.

src/apps/FEMSWEEP.cpp
```diff
@@ -152,132 +123,66 @@ void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type t
...
-                      1*sizeof(Index_type) * 10800 + // F_g2l
-                      1*sizeof(Index_type) * 37800 + // idx1
-                      1*sizeof(Index_type) * 37800 );// idx2
-  setBytesWrittenPerRep( 0 );
-  setBytesModifyWrittenPerRep( 1*sizeof(Real_type) * m_Xlen ); // Xdat
+                      1*sizeof(Index_type) * (m_sharedinteriorfaces + m_boundaryfaces) + // F_g2l
+                      1*sizeof(Index_type) * m_sharedinteriorfaces * 4 + // idx1
+                      1*sizeof(Index_type) * m_sharedinteriorfaces * 4 );// idx2
+  Index_type amount_Xdat_modify_written = 1.55 * m_na * m_ng * m_sharedinteriorfaces;
+  setBytesWrittenPerRep( 1*sizeof(Real_type) * (m_Xlen - amount_Xdat_modify_written) ); // remainder of Psi after sharedinteriorfaces
```
MrBurmark: I believe this is called Xdat here

src/apps/FEMSWEEP.cpp
```diff
@@ -152,132 +123,66 @@ void FEMSWEEP::setSize(Index_type RAJAPERF_UNUSED_ARG(target_size), Index_type t
...
+  setBytesModifyWrittenPerRep( 1*sizeof(Real_type) * amount_Xdat_modify_written );  // a multiple of sharedinteriorfaces
```
MrBurmark:
```suggestion
  setBytesModifyWrittenPerRep( 1*sizeof(Real_type) * amount_Xdat_modify_written );  // Xdat, estimate as this depends on mesh and angles
```
How about something more like this?

src/apps/FEMSWEEP.hpp
```diff
@@ -10,53 +10,54 @@
...
 ///   // number and offset of hyperplanes for this angle
-///   const int nhp = nhpaa_r[a];
-///   const int ohp = ohpaa_r[a];
+///   const Index_type nhp = nhpaa_r[a];
+///   const Index_type ohp = ohpaa_r[a];
 ///   // elements in this hyperplanes processed so far
-///   int s_nehp_done = 0;
-///   double A[ND * ND], b[ND];
+///   Index_type s_nehp_done = 0;
+///   double A[ND*ND];
```
MrBurmark: Real_type?

src/apps/FEMSWEEP.hpp
```diff
@@ -119,74 +122,79 @@ constexpr int FDS = 4;  // number of DOFs per face
...
-  int s_nehp_done = 0; \
-  double A[ND * ND], b[ND]; \
-  double Ffactor = fmax(sin(Adat[order_r[a*ne]*ND*ND + a*ne*ND*ND]) - 2.0, 0.0); \
-  for (int hp = 0; hp < nhp; ++hp) \
+  const Index_type a = ag / ng; \
+  const Index_type g = ag % ng; \
+  const Index_type nhp = nhpaa_r[a]; \
+  const Index_type ohp = ohpaa_r[a]; \
+  Index_type s_nehp_done = 0; \
+  double A[ND*ND]; \
```
MrBurmark: Real_type

src/apps/FEMSWEEP.hpp
```diff
@@ -119,74 +122,79 @@ constexpr int FDS = 4;  // number of DOFs per face
...
 template <int N>
-RAJA_HOST_DEVICE inline void SolveLinearSystemNxN(double *A, 
-                                                  const double s,
-                                                  const double *M, 
-                                                  const double *b, 
-                                                  double * x)
+RAJA_HOST_DEVICE inline void SolveLinearSystemNxN(Real_ptr A, 
+                                                  const Real_type s,
+                                                  Real_const_ptr M, 
+                                                  const double * b,
```
MrBurmark: Real_type

src/common/RunParams.cpp
```diff
@@ -699,6 +710,103 @@ void RunParams::parseCommandLineOptions(int argc, char** argv)
...
+                  << std::endl;
+        input_state = BadInput;
+      }
+
+    } else if ( opt == std::string("--femsweep_groups") ) {
+
+      i++;
+      if ( i < argc ) {
+        long long num = ::atoll( argv[i] );
+        long long min_num = 1;
```
artv3, line 760: Do these need RAJA types?

rchen20: This is just for parsing input. Not sure if it's worth changing all of the input parsing for our RAJA types.

PR #656, opened 2026-03-17
Scripts for a compiler matrix

artv3:
# Summary

- This PR is adds helper scripts to do cross compiler performance comparisons 

Sample run:
See: data/compiler_sweeps/tier1_base_raja_hip_repro.txt

Folks, these scripts are AI generated. But I spent time looking over the code and it looks solid to me. I would like to invite the team to review.


scripts/lc-builds/run_compiler_matrix.sh
```diff
@@ -0,0 +1,352 @@
+#!/usr/bin/env bash
```
rhornung67, line 1: I think this script should have a more descriptive name. Not sure what it could be, but worth thinking a bit about.

artv3:
Consider that we may want different MPI variants

artv3:
Consider throughput comparisons for different compilers

artv3:
To compare compiler performance consider putting together a ratio plot between a base compiler

scripts/lc-builds/run_compiler_matrix.sh
artv3: Hi @rhornung67 , what if we call it the compiler matrix, since that is what we are building?

scripts/lc-builds/run_compiler_matrix.sh
```diff
@@ -0,0 +1,569 @@
...
+# to contribute to RAJA Performance Suite.
+#
+# SPDX-License-Identifier: (BSD-3-Clause)
+###############################################################################
+
+set -uo pipefail
+
+usage() {
+  cat <<'EOF'
+Usage:
```
johnbowen42, line 23: Is there a reason to prefer a shell script over a python script here?  I thought maybe it would dovetail better with our burgeoning analysis scripts to do Python

scripts/lc-builds/run_compiler_matrix.sh
```diff
@@ -0,0 +1,569 @@
...
+      ;;
+    --fail-fast)
+      keep_going=0
+      shift
+      ;;
+    --log-dir)
+      log_dir="${2:-}"
+      shift 2
+      ;;
+    -h|--help)
```
johnbowen42, line 231: i.e. we could do an argparse here instead of this

scripts/lc-builds/run_compiler_matrix.sh
```diff
@@ -0,0 +1,569 @@
...
+    fi
+  else
+    entry_build_script="${build_script}"
+    script_args=("${line_tokens[@]}")
+  fi
+  log_prefix="${log_dir}/${entry_id}"
+
+  echo "==> [${entry_id}] line ${line_no}: $(basename -- "${entry_build_script}") ${script_args[*]}"
+
+  # Execute each matrix entry in a fresh login shell so module operations and
```
johnbowen42, line 391: Is this why it's easier to handle this with a shell script?

rhornung67:
@artv3 would you please add a short section to the RAJAPerf User Guide explaining how to use this?

docs/sphinx/user_guide/run.rst
```diff
@@ -97,7 +97,7 @@ generated in the run directory after Suite execution completes.
 .. note:: You can pass the ``--dryrun`` command-line option to the executable
           to see a summary of how the Suite will execute by showing run
           parameters without actually running it. You can pass any other
-          command-line options when doing a "dry run" and you will see
+          command-line options when doing a "dryrun" and you will see
```
artv3:
```suggestion
          command-line options when doing a "dry run" and you will see
```

scripts/lc-builds/run_compiler_matrix.sh
artv3: My thoughts are that since there the workflow is environment dependent, it would be more fitting to use a shell script. For example its consistent with the lc-build scripts, handles stuff like `srun, cmake, module` a bit more naturally

scripts/lc-builds/run_compiler_matrix.sh
artv3: yes

docs/sphinx/user_guide/run.rst
```diff
@@ -130,6 +130,35 @@ in the Suite.
...
+``srun`` to stop parsing its own options so the matrix script can append
+``./bin/raja-perf.exe`` and the RAJAPerf run arguments after it.
+
+The compiler list file contains one build configuration per non-comment line.
+If you pass ``from-list`` as the build-script argument, the first token on each
+line names the build script and the remaining tokens are passed to that script.
+
+The generated ``*-kernel-run-data.csv`` files can then be merged and plotted
+with the processing scripts in ``scripts/`` such as
+``scripts/build_compiler_matrix.py`` and ``scripts/kh_compiler_matrix.py``.
```
artv3:
```suggestion
``scripts/build_compiler_matrix.py`` where it will generate corresponding plots.
```

MrBurmark:
@artv3 I updated the script to also output a basic html landing page and per kernel html pages in addition to what it was outputting before. This makes it easier to navigate through the output files.
I also change the plots to be log-log.

scripts/lc-builds/toss4_mvapich2_nvcc_gcc.sh
```diff
@@ -31,6 +31,14 @@ NVCC_COMP_VER=$2
 NVCC_COMP_ARCH=$3
 GCC_COMP_VER=$4
 
+# Pick a default C++ standard compatible with the CUDA toolchain.
+# CUDA 11.x does not support C++20 in NVCC mode (BLT enforces this).
+NVCC_MAJOR="${NVCC_COMP_VER%%.*}"
+BLT_CXX_STD="c++17"
```
MrBurmark: Not sure this is relevant anymore.

rhornung67: No, it is not. We can't use CUDA < 12 when we require C++20

artv3:
@llnl/raja-core are we good to go?

docs/sphinx/user_guide/run.rst
```diff
@@ -106,7 +106,7 @@ a summary of command-line arguments it was given if the input contains
 something that the code does not know how to parse. For example, running the
 command::
 
-  $ ./bin/raja-perf.exe --dry-run -k DAXPY Foo
+  $ ./bin/raja-perf.exe --dryrun -k DAXPY Foo
```
rhornung67, line 109: Thank you for fixing this

docs/sphinx/user_guide/run.rst
```diff
@@ -130,6 +130,44 @@ in the Suite.
...
+
+For compiler sweeps, use ``scripts/lc-builds/run_compiler_matrix.sh`` rather
+than invoking a single build script by hand. The matrix driver reads a compiler
+list file, sources the selected build script for each line, builds each
+configuration, and then runs the Suite with the launcher command you provide
+via ``--run-cmd``.
+
+A typical invocation looks like::
+
+  $ ./scripts/lc-builds/run_compiler_matrix.sh toss4_amdclang.sh compiler_list.txt \
```
rhornung67, line 148: Why is `the toss4_amdclang.sh` argument included? Is it added to contents of `compiler_list.txt` file?

In other words, does the example command show that compilers can be specified by listing scripts explicitly on the command line as well as listing them in a file?

rhornung67 (approved):
I asked some questions in the comments.

data/compiler_sweeps/tier1_base_raja_hip_repro.txt
```diff
@@ -0,0 +1,24 @@
+Compiler matrix input:
```
MrBurmark, line 1: Would this be better as a markdown file?

MrBurmark (approved):

rhornung67:
@artv3 this is approved, but there are a few questions from @MrBurmark and @rhornung67. Please consider those before merging.

artv3:
> @artv3 this is approved, but there are a few questions from @MrBurmark and @rhornung67. Please consider those before merging.

Sure thing! I'll push up a clean up pass based on the feedback/suggestions

docs/sphinx/user_guide/run.rst
artv3: One way to think about it is that toss4_amdclang.sh selects the build-script template (held fixed for the whole run), and compiler_list.txt provides one set of arguments per build (one entry per non-comment line) that gets passed to that script. This makes it so everything has the same build flags. Does that make sense?

data/compiler_sweeps/tier1_base_raja_hip_repro.txt
artv3: pushed up

docs/sphinx/user_guide/run.rst
rhornung67: I think it would be good to explain these details in the user docs.

artv3: Sounds good, I pushed up additional details on things work and the different ways to build. If it looks reasonable I can merge.

rhornung67 (approved):
Thank you for addressing my comments @artv3

PR #659, opened 2026-03-19
FEMSWEEP hyperplane parallelism

MrBurmark:
# Summary

Add parallelism within each hyperplane for GPU backends.
Uncollapse angles and groups loops. Note that launch does not support omp collapse so RAJA_OpenMP is still manually collapsed.

Also fix the way mesh sizes are picked for a target problem size in FEMSWEEP::sizeSize.

- This PR is a feature
- It does the following:
  - Adds optimization to FEMSWEEP kernel implementations at the request of me

### DIFF
--- a/src/apps/FEMSWEEP.cpp
+++ b/src/apps/FEMSWEEP.cpp
@@ -28,10 +28,15 @@ namespace apps
 FEMSWEEP::FEMSWEEP(const RunParams& params)
   : KernelBase(rajaperf::Apps_FEMSWEEP, params)
 {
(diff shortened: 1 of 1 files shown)


src/apps/FEMSWEEP-OMP.cpp
```diff
@@ -13,6 +13,11 @@
 
 #include <iostream>
 
+
+#define USE_OMP_COLLAPSE
+//#undef USE_OMP_COLLAPSE
```
rchen20, line 18: Should we remove this undef?

MrBurmark: I agree this is weird, but I did this as I was following a pattern in other files.

PR #680, opened 2026-04-15
Feature(Kokkos-PolyBench): Add Jacobi 1D/2D and Heat 3D Kokkos implementations

Adrien-Tab:
## Summary
Add a directory for kokkos-polybench with 3 benchmarks.

- This PR is a feature
- It adds the Kokkos version of the following PolyBench suites:
  - Jacobi_1D
  - Jacobi_2D
  - Heat_3D

PR #681, opened 2026-04-16
Pr from fork/680

MrBurmark:
# Summary

- This PR is a copy of #680 that can run CI and be merged

### DIFF
--- a/src/polybench-kokkos/POLYBENCH_HEAT_3D-Kokkos.cpp
+++ b/src/polybench-kokkos/POLYBENCH_HEAT_3D-Kokkos.cpp
@@ -23,7 +23,6 @@ void POLYBENCH_HEAT_3D::runKokkosVariant(VariantID vid) {
 
   POLYBENCH_HEAT_3D_DATA_SETUP;
 
-  // Views 3D : dimension order (i, j, k), i.e. (N, N, N)
   auto A_view = getViewFromPointer(A, N, N, N);
   auto B_view = getViewFromPointer(B, N, N, N);
 
--- a/src/polybench-kokkos/POLYBENCH_HEAT_3D-Kokkos.cpp
+++ b/src/polybench-kokkos/POLYBENCH_HEAT_3D-Kokkos.cpp
@@ -1,0 +1,91 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) Lawrence Livermore National Security, LLC and other
+// RAJA Project Developers. See top-level LICENSE and COPYRIGHT
+// files for dates and other details. No copyright assignment is required
+// to contribute to RAJA Performance Suite.
+//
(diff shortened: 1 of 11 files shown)


src/polybench-kokkos/POLYBENCH_HEAT_3D-Kokkos.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+namespace rajaperf {
+namespace polybench {
+
+void POLYBENCH_HEAT_3D::runKokkosVariant(VariantID vid) {
+  const Index_type run_reps = getRunReps();
+
+  POLYBENCH_HEAT_3D_DATA_SETUP;
+
+  // Views 3D : dimension order (i, j, k), i.e. (N, N, N)
```
artv3: Does Kokkos follow the C-style index ordering? Fastest stride is the right most index? 
Wondering if its worth noting?

Adrien-Tab: No, it depends on the configured backend. I think this comment can be removed.

artv3: @Adrien-Tab can you add a comment specifying which ordering is in effect?

artv3: The other thought I had was if kokkos had a backend dependent ordering stride is that worth also having with the native and RAJA variants? This would promote consistency if it is the case. Or have Kokkos use the same striding order used by native and RAJA again for consistency regardless of backend.

MrBurmark: Kokkos should use the same data layout as the other variants of the kernel so the results are as comparable across variants. There are multi dimensional views in multiple files, let's not worry about documenting or fixing that in this PR if it needs to be fixed.

MrBurmark:
```suggestion
```

artv3: Sounds good, please capture in an issue if there is an inconsistency.

MrBurmark: @Adrien-Tab are there any cases where kokkos is reordering the data?

rhornung67 (approved):

src/polybench-kokkos/POLYBENCH_HEAT_3D-Kokkos.cpp
Adrien-Tab: I'm not sure about the reordering question. The default layout shouldn't be an issue since the `View` default layout and the parallel pattern default iteration order are matched per backend. For GPU backends this means `LayoutLeft` (Fortran-style indexing).

artv3: Ah interesting, RAJA's default layout is `LayoutRight` C-style indexing.

PR #683, opened 2026-05-20
[WIP] Add Branson Monte Carlo Kernel

michaelmckinsey1:
# Summary
An intern (@RanivG) will be implementing https://github.com/lanl/branson/blob/3f538e31a51b144b6621d4665f465d8d44ee85b8/src/event_based_transport.h#L323 as a kernel in RAJAPerf.

- [ ] Add `Apps_MC_HISTORY_PARTICLE_TRANSPORT` kernel
- [x] Add reference implementation to header file (see other kernels for examples)
- [ ] CPU-only implementation
- [ ] RAJA implementation
- [ ] CUDA implementation
- [ ] HIP implementation
- [x] Rename to `MC_HISTORY_PARTICLE_TRANSPORT`


MrBurmark:
Is there a better name for this kernel?

michaelmckinsey1:
> Is there a better name for this kernel?

I am open to suggestions, but it can be renamed easily.

rhornung67:
I'm fine with the name. @MrBurmark do you have a specific concern about the name?

MrBurmark:
The name is just very generic. Is the kernel any specific part of the Monty Carlo algorithm?

rhornung67:
> The name is just very generic. Is the kernel any specific part of the Monty Carlo algorithm?

That's a good point. A brief scan of the code at the link @michaelmckinsey1 provided above to the Branson code, perhaps something like MC_EVENT_PHOTON_TRANSPORT, if that's not too long.

rhornung67:
@michaelmckinsey1 please add a reference implementation to the kernel header file.

dane7-blackhole-1516011.core
artv3, line 1: probably don't want this and would be good to double check on the other extra files

rhornung67: Yes, there are a bunch of unwanted files committed in this. Please refrain from using `git add *`.

src/apps/TRANSPORT3DMC-Seq.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+#if defined(RUN_RAJA_SEQ)
+    case Lambda_Seq : {
+
+      startTimer();
+      transport3dmc_lam_setup(vid, iend);
+      // Loop counter increment uses macro to quiet C++20 compiler warning
+      for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
+        transport3dmc_lam_reset();
```
MrBurmark: Only the loop body needs to be in a lambda.

src/apps/TRANSPORT3DMC-Seq.cpp
```diff
@@ -0,0 +1,107 @@
...
+                                    TRANSPORT3DMC_RESET
+                                  };
+#endif
+
+  switch ( vid ) {
+
+    case Base_Seq : {
+
+      startTimer();
+      TRANSPORT3DMC_DATA_SETUP(vid, iend);
```
MrBurmark: Setup should happen once above the variant switch statement.

src/apps/TRANSPORT3DMC-Seq.cpp
```diff
@@ -0,0 +1,107 @@
...
+      // Loop counter increment uses macro to quiet C++20 compiler warning
+      for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
+        TRANSPORT3DMC_RESET
```
MrBurmark: If reset involves a loop from begin to end, you probably want to express that loop outside of the reset macro.

src/apps/TRANSPORT3DMC-Seq.cpp
MrBurmark: Unless the reset is also a loop body.

src/apps/TRANSPORT3DMC.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+  setItsPerRep( getActualProblemSize() );
+  setKernelsPerRep(1);
+
+  setBytesAllocatedPerRep( 0 ); // in, out
+  setBytesReadPerRep( 0 ); // in
+  setBytesWrittenPerRep( 0 ); // out
+  setBytesModifyWrittenPerRep( 0 );
+  setBytesAtomicModifyWrittenPerRep( 0 );
+  setFLOPsPerRep( 0 );
```
MrBurmark: Might want to add a TODO comment as a reminder.

src/apps/TRANSPORT3DMC.cpp
```diff
@@ -0,0 +1,433 @@
...
+    binUpperBound += binWidth;
+    bins[i] = binUpperBound;
+  }
+  CALI_MARK_END("Bin Init");
+
+}
+
+void TRANSPORT3DMC::initMaterials(std::vector<TRANSPORT3DMC::Material> &mats) {
+  size_t matCt = 12;
+  mats.resize(matCt);
```
MrBurmark: You'll probably want to use something like `allocDataForInit` to get memory here and allow you to manually set it up. Note that all setup for the data must be completed before the object returned by `allocDataForInit` goes out of scope, see examples by searching for `ForInit`.

src/apps/TRANSPORT3DMC.cpp
```diff
@@ -0,0 +1,433 @@
...
+  setBytesModifyWrittenPerRep( 0 );
+  setBytesAtomicModifyWrittenPerRep( 0 );
+  setFLOPsPerRep( 0 );
+}
+
+TRANSPORT3DMC::~TRANSPORT3DMC()
+{
+}
+
+int TRANSPORT3DMC::Cell::GLBL = 0;
```
MrBurmark: Why are you using static class variables?

src/apps/TRANSPORT3DMC.hpp
```diff
@@ -0,0 +1,293 @@
...
+  void tearDown(VariantID vid, size_t tune_idx);
+
+  void defineSeqVariantTunings();
+
+  void runSeqVariant(VariantID vid);
+
+  #define dT 0.001
+  #define N_ISOTOPE 68
+  //#define GROUPS 10
+  #define cutoff 1e-5
```
MrBurmark: Defs should be above.

MrBurmark: Though you may be able to make these `static constexpr` class variables instead.

src/apps/MC_HISTORY_PARTICLE_TRANSPORT.hpp
```diff
@@ -0,0 +1,293 @@
...
+    XS(std::mt19937_64 &RNGr, std::uniform_real_distribution<double> &posDist);
+    XS operator*(const double c) const;
+    XS operator+(const XS &sigma) const;
+  };
+
+  struct Cell {
+    public:
+      static int GLBL;
+      int ID;
+      std::array<double, 6> planes; //-x, +x, -y, +y, -z, +z
```
MrBurmark, line 245: You might want to use camp::array, its the same as std::array but works on gpus.

src/apps/MC_HISTORY_PARTICLE_TRANSPORT.hpp
```diff
@@ -0,0 +1,293 @@
...
+  };
+
+  enum BC {
+    ELEMENT,
+    REFLECT,
+    VACUUM,
+    SOURCE,
+  };
+
+  struct XS {
```
MrBurmark: Rename to `CrossSection` for clarity?

src/apps/TRANSPORT3DMC.hpp
```diff
@@ -0,0 +1,293 @@
...
+    double scatter;
+    double abs;
+    double fission;
+    double nu_fission;
+    double total;
+
+    XS();
+    XS(std::mt19937_64 &RNGr, std::uniform_real_distribution<double> &posDist);
+    XS operator*(const double c) const;
+    XS operator+(const XS &sigma) const;
```
MrBurmark: Define inline for performance?

src/apps/TRANSPORT3DMC.hpp
```diff
@@ -0,0 +1,293 @@
...
+      std::array<int, 6> next;
+      int matID;
+
+      Cell(); 
+      double getBoundary(const std::array<double,3> &pos, const std::array<double,3> &angle, uint32_t &surface_cross);
+  };
+
+  struct Material {
+    int isotopeCt;
+    std::vector<double> conc;
```
MrBurmark: Prefer `allocData` and `deallocData` and related helper functions to get memory.

src/common/RunParams.cpp
```diff
@@ -47,6 +47,9 @@ RunParams::RunParams(int argc, char** argv)
    ltimes_num_d(6),
    ltimes_num_g(32),
    ltimes_num_m(25),
+   transport3dMC_particles(10000),
+   transport3dMC_cube_dim(3),
+   transport3dMC_groups(10),
```
MrBurmark: Nice, we'll have to have to think about what are reasonable defaults later.

.gitignore
```diff
@@ -14,3 +14,4 @@
 /install_*/
 /install-*/
 /Debug/
+*.cmake
```
MrBurmark: This seems a bit broad, don't we use cmake?

src/apps/MC_HISTORY_PARTICLE_TRANSPORT.cpp
```diff
@@ -0,0 +1,433 @@
...
+{
+
+
+TRANSPORT3DMC::TRANSPORT3DMC(const RunParams& params)
+  : KernelBase(rajaperf::Apps_TRANSPORT3DMC, params)
+{
+  dim = params.getTransport3DmcCubeSz();
+  GROUPS = params.getTransport3DmcGroups();
+  setDefaultProblemSize(1);
+  setDefaultReps(1000);
```
MrBurmark, line 28: This is fine for now, long term I'm going to guess this is going to move toward values like this.

```suggestion
  setDefaultProblemSize(1000000);
  setDefaultReps(10);
```

src/apps/TRANSPORT3DMC.cpp
RanivG: Since the mesh is an array, the static GLBL helps to identify/assign cells their ID/index into the mesh vector. Can probably be removed once all vectors have been migrated to pointers.

src/apps/MC_HISTORY_PARTICLE_TRANSPORT.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+  setSize(params.getTargetSize(getDefaultProblemSize()),
+          params.getReps(getDefaultReps()));
+
+  setChecksumConsistency(ChecksumConsistency::ConsistentPerVariantTuning);
```
MrBurmark, line 33: Hopefully this stays consistent as we'll have consistent pseudo random numbers.

src/apps/TRANSPORT3DMC.cpp
```diff
@@ -0,0 +1,433 @@
...
+  
+  return pState;
+}
+
+//Material constructor function
+TRANSPORT3DMC::Material::Material() {
+  isotopeCt = 1;
+  conc = std::vector<double>();
+  nucIDs = std::vector<int>();
+}
```
MrBurmark: This is fine for now, but c++ has a syntax for initializing class members https://en.cppreference.com/cpp/language/constructor.

RanivG: Ended up just deleting it and letting the compiler define a default constructor.

.gitignore
RanivG: removed, checked out the gitignore from develop

src/apps/TRANSPORT3DMC.hpp
RanivG: implemented.

src/apps/TRANSPORT3DMC.hpp
RanivG: fixed

src/apps/TRANSPORT3DMC.cpp
RanivG: implemented

src/apps/TRANSPORT3DMC-Seq.cpp
RanivG: No loop, just a deep copy of the reference set to another object actually conducting transport.

src/apps/TRANSPORT3DMC-Seq.cpp
RanivG: fixed

src/apps/TRANSPORT3DMC-Seq.cpp
RanivG: fixed

dane7-blackhole-1516011.core
RanivG: fixed.

RanivG:
I was reading through earlier comments, I disagree with the name MC_EVENT_PHOTON_TRANSPORT; the kernel emulates history-based transport of neutrons (or at least that's how I designed it). 
Might I propose "HISTORY_TRANSPORT3DMC" ?

michaelmckinsey1:
> I was reading through earlier comments, I disagree with the name MC_EVENT_PHOTON_TRANSPORT; the kernel emulates history-based transport of neutrons (or at least that's how I designed it). Might I propose "HISTORY_TRANSPORT3DMC" ?

If the difference is history vs event, why not `MC_HISTORY_PHOTON_TRANSPORT` then?

RanivG:
> > I was reading through earlier comments, I disagree with the name MC_EVENT_PHOTON_TRANSPORT; the kernel emulates history-based transport of neutrons (or at least that's how I designed it). Might I propose "HISTORY_TRANSPORT3DMC" ?
> 
> If the difference is history vs event, why not `MC_HISTORY_PHOTON_TRANSPORT` then?

Well it's not really photons, at least I don't think. Also I was more trying to keep with the 3D kernel's naming scheme (MASS3D, CONVECTION3D, VOL3D, DIFFUSION3D, etc). If this convention isn't desired for this kernel then we can stay with the `MC_HISTORY_PARTICLE_TRANSPORT` convention

MrBurmark:
> Well it's not really photons, at least I don't think. Also I was more trying to keep with the 3D kernel's naming scheme (MASS3D, CONVECTION3D, VOL3D, DIFFUSION3D, etc). If this convention isn't desired for this kernel then we can stay with the `MC_HISTORY_PARTICLE_TRANSPORT` convention

I don't think there is much reason to limit yourself to that 3D naming convention as this is a different type of kernel.

build-linux-rhel8-sapphirerapids-5szrqd5
michaelmckinsey1, line 1: @RanivG Rm this

build-linux-rhel8-sapphirerapids-yuxnhrq
michaelmckinsey1, line 1: @RanivG Rm this

RanivG: removed

PR #684, opened 2026-06-10
Add LTimes tuning to represent Kripke LTimes

michaelmckinsey1:
Matching parameters like zones, groups, moments, and directions, the runtime of RAJAPerf LTimes is equivalent (<1%) to Kripke LTimes on CPU. But I have noticed runtime differences on GPU by ~15% on CUDA and ~60% on ROCm. After these changes CUDA runtime is within 1% and ROCm within 6% (CUDA is now faster, ROCm actually (~60%) _slower_ with block 25). This is because AMD wavefront size 64 on block 25 wastes significantly more threads than NVIDIA warp size 32.

# Summary

- This PR is a refactoring
  - [x] Modifies `ell` layout from `d` contiguous to `m` contiguous to make accesses coalesced. This matches Kripke ZGD, where `Field_Ell` is layout ordering `moment` as stride-1. Doesn't apply to psi because does not depend on m (threads access same value), phi because no d (already contiguous in `m`).
     - This change made ~5% runtime difference on CUDA
  - [x] Adds a new tuning `block_m`/`kernel_m` where the default block size to be `m` instead of 256 and remapped the kernel so z -> blockIdx.x, g -> blockIdx.y, m -> threadIdx.x. This changes the launch to `grid=(num_z, num_g, 1), block=(m, 1, 1)`, which matches the kripke launch. However, running at different legendre orders (`m`) will result in different block sizes.
  - [x] ~Would need to make similar changes for `LTIMES-NOVIEW`?~ done
  - [x] Improve tuning names
     - New tuning: `kernel_m_{blocksize}`
     - Existing tuning: `kernel_zgm_{blocksize}`


MrBurmark:
Yes you would have to make LTIMES and LTIMES_NO_VIEW match.

src/apps/LTIMES.hpp
MrBurmark: Did you want to have m be stride-1 here?

src/apps/LTIMES.hpp
```diff
@@ -76,7 +78,7 @@ namespace ltimes_idx {
                                    RAJA::Layout<3, Index_type, 2>,
                                    IZ, IG, ID>;
   using ELL_VIEW = RAJA::TypedView<Real_type,
-                                   RAJA::Layout<2, Index_type, 1>,
+                                   RAJA::Layout<2, Index_type, 0>,
                                    IM, ID>;
```
MrBurmark, line 82: This only tells the layout which dimension will be stride-1, it doesn't change how the layout is constructed to enforce that this is true.

src/apps/LTIMES.hpp
michaelmckinsey1: Yes fixed

src/apps/LTIMES.hpp
michaelmckinsey1: Screenshot 2026-06-18 at 2 38 15 PM

After fixing the layout, I now verified that the checksum matches before and after these changes (the changes are shown in reverse here).

michaelmckinsey1:
> Yes you would have to make LTIMES and LTIMES_NO_VIEW match.

Done

src/apps/LTIMES.hpp
artv3: For views, I'm a fan of using the ``RAJA::make_permuted_view`` method much simpler I think: 
https://github.com/llnl/RAJA/blob/develop/examples/make_permuted_view.cpp#L51

artv3: It takes care of the fast index by selecting either ``RAJA::layout_{right,left}``

src/apps/LTIMES-Cuda.cpp
```diff
@@ -139,10 +138,10 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
...
       using EXEC_POL =
         RAJA::KernelPolicy<
-          RAJA::statement::CudaKernelFixedAsync<m_block_sz*g_block_sz*z_block_sz,
-            RAJA::statement::For<1, RAJA::cuda_global_size_z_direct<z_block_sz>,     //z
-              RAJA::statement::For<2, RAJA::cuda_global_size_y_direct<g_block_sz>,   //g
-                RAJA::statement::For<3, RAJA::cuda_global_size_x_direct<m_block_sz>, //m
+          RAJA::statement::CudaKernelAsync<
+            RAJA::statement::For<1, RAJA::cuda_block_x_loop, // z
+              RAJA::statement::For<2, RAJA::cuda_block_y_loop, // g
+                RAJA::statement::For<3, RAJA::cuda_thread_x_loop, // m
```
MrBurmark, line 212: I assume this is a non-size loop policy because it is in ltimes. Here we know the block size at compile time, is that also true in kripke?

src/apps/LTIMES_NOVIEW.hpp
```diff
@@ -86,9 +86,9 @@ class LTIMES_NOVIEW : public KernelBase
   void runSyclVariantImpl(VariantID vid);
 
 private:
-  static const size_t default_gpu_block_size = 256;
-  using gpu_block_sizes_type = integer::make_gpu_block_size_list_type<default_gpu_block_size,
-                                                         integer::MultipleOf<32>>;
+  static const size_t default_gpu_block_size = 25;
```
rchen20: How was a block size of 25 chosen? Would it be better for both GPU platforms to set this to 32?

src/apps/LTIMES-Cuda.cpp
```diff
@@ -139,10 +138,10 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
 
       using EXEC_POL =
         RAJA::KernelPolicy<
-          RAJA::statement::CudaKernelFixedAsync<m_block_sz*g_block_sz*z_block_sz,
-            RAJA::statement::For<1, RAJA::cuda_global_size_z_direct<z_block_sz>,     //z
-              RAJA::statement::For<2, RAJA::cuda_global_size_y_direct<g_block_sz>,   //g
-                RAJA::statement::For<3, RAJA::cuda_global_size_x_direct<m_block_sz>, //m
+          RAJA::statement::CudaKernelAsync<
```
rchen20, line 209: Note that I've reverted LTimes to launch synchronously in Kripke, for correctness. This is fine though because the direction loop is inner-most, which should avoid race conditions.

src/apps/LTIMES-Cuda.cpp
rchen20: In Kripke, the block sizes are determined by parameters passed in at runtime. For example, like in this version of LTimes, it will be blocked on zones and groups. The exact parameters which will be blocked are not always the same for each Kripke run. For instance, if we use the DZG layout at runtime, then the loops will be blocked with directions and zones (while groups are threaded).

MrBurmark:
Do we want to only have this kripke conforming tuning, or should does it make sense to keep the current tuning as well?

artv3:
@michaelmckinsey1 take a look here: https://github.com/llnl/RAJA/blob/develop/benchmark/ltimes.cpp, it would be cool to also have a GPU shared memory version as a tuning!

src/apps/LTIMES_NOVIEW.hpp
michaelmckinsey1: It is the square of `legendre order + 1`. In kripke this is also the case, the default legendre is 4, so (4+1)^2 means the kernel in kripke will be `(5,5,1)`

In RAJAPerf, we set `m=25` directly (and this is the default). So `m=36` would then be equivalent to legendre=5 in kripke.

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: It has been async in RAJAPerf, I just changed it from `CudaKernelFixedAsync` to `CudaKernelAsync`, but for completeness I can make it `CudaKernel`. I don't this would matter for performance in RAJAPerf.

src/apps/LTIMES-Cuda.cpp
```diff
@@ -172,31 +246,114 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
...
+
+      startTimer();
+      // Loop counter increment uses macro to quiet C++20 compiler warning
+      for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
+
+        RAJA::launch<launch_policy>( res,
+            RAJA::LaunchParams(RAJA::Teams(*num_z, *num_g, 1),
+                               RAJA::Threads(block_size, 1, 1)),
+            [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+
```
artv3, line 298: It might be worth looking into the CachePolicy for the launch_context: 

https://github.com/llnl/RAJAPerf/blob/develop/src/apps/MASS3DEA-Cuda.cpp#L206-L209

We got some speedups here, this might help the nested loops

michaelmckinsey1:
> Do we want to only have this kripke conforming tuning, or should does it make sense to keep the current tuning as well?

I have now Implemented the Kripke implementation as a separate tuning, since we see many cases where the kripke tuning is slower.

src/apps/LTIMES-Cuda.cpp
```diff
@@ -242,12 +399,35 @@ void LTIMES::defineCudaVariantTunings()
...
+  for (VariantID vid : {Base_CUDA, Lambda_CUDA, RAJA_CUDA}) {
+
+    seq_for(factorized_gpu_block_sizes_type{}, [&](auto block_size) {
+
+      if (run_params.numValidGPUBlockSize() == 0u ||
+          run_params.validGPUBlockSize(block_size)) {
+
+        if (vid == RAJA_CUDA) {
+          addVariantTuning<&LTIMES::runCudaVariantImpl<block_size, 1>>(
+              vid, "kernel_m32_gBigFact_zLowFact_"+std::to_string(block_size));
```
michaelmckinsey1: @MrBurmark Do you think `kernel_{blocksize}_factorized` would be a better name?

And then maybe the Kripke version becomes `kernel_{blocksize}_m`

MrBurmark: How about something like `block_m_{blocksize}` and `block_gzm_{blocksize}` or even `block_g{g_block_size}_z{z_blocksize}_m{m_blopcksize}`?

michaelmckinsey1: kripke tuning: `block_m_{blocksize}`
original tuning: `block_gzm_{blocksize}` or `block_g{g_block_size}_z{z_blocksize}_m{m_blocksize}` both work, but I prefer `block_gzm_{blocksize}`

MrBurmark: Ya shorter makes sense.

MrBurmark: I'm not sure that gzm is the order actually used in the block or not, but whatever is used in the loop in the right order.

michaelmckinsey1: it's zgm

michaelmckinsey1: Tuning names are now updated to `kernel_m_{blocksize}` and `kernel_zgm_{blocksize}`

src/apps/LTIMES-Cuda.cpp
MrBurmark: The size policies tell the implementation the compile time dimensions of the block and/or grid, not the extents of the iteration space.
```suggestion
                RAJA::statement::For<3, RAJA::cuda_thread_size_x_loop<block_size>, // m
```

src/apps/LTIMES-Cuda.cpp
```diff
@@ -27,27 +27,60 @@ using namespace ltimes_idx;
 //
 // Define thread block shape for CUDA execution
 //
-#define m_block_sz (32)
-#define g_block_sz (integer::greater_of_squarest_factor_pair(block_size/m_block_sz))
-#define z_block_sz (integer::lesser_of_squarest_factor_pair(block_size/m_block_sz))
+#define block_moments_m_block_sz (block_size)
```
MrBurmark: Do you want to change these macro names to more closely match the tuning names? Like `m_block_sz` as that tuning has a 1D block and `zgm_m_block_sz`, `zgm_g_block_sz`, `zgm_m_block_sz` as that tuning has a 3D block.

src/apps/LTIMES-Cuda.cpp
```diff
@@ -118,17 +179,30 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
...
       constexpr size_t shmem = 0;
 
-      RPlaunchCudaKernel(
-        (ltimes_lam<LTIMES_THREADS_PER_BLOCK_TEMPLATE_PARAMS_CUDA,
-                    decltype(ltimes_lambda)>),
-        nblocks, nthreads_per_block,
-        shmem, res.get_stream(),
-        num_m, num_g, num_z,
-        ltimes_lambda );
+      if constexpr (tune_idx == 1) {
```
MrBurmark: Use tune_idx in order consistently.

src/apps/LTIMES.hpp
```diff
@@ -114,17 +116,21 @@ class LTIMES : public KernelBase
...
   template < size_t work_group_size, size_t tune_idx >
   void runSyclVariantImpl(VariantID vid);
 
 private:
-  static const size_t default_gpu_block_size = 256;
-  using gpu_block_sizes_type = integer::make_gpu_block_size_list_type<default_gpu_block_size,
-                                                         integer::MultipleOf<32>>;
+  static const size_t default_gpu_block_size = 25;
+  using gpu_block_sizes_type =
+      integer::make_gpu_block_size_list_type<default_gpu_block_size>;
```
MrBurmark: If there is only one block size for m we probably don't need to make a list. If that block size is only known at runtime then there is no need to have it here.

src/apps/LTIMES.hpp
```diff
@@ -114,17 +116,21 @@ class LTIMES : public KernelBase
...
   void runOpenMPTargetVariant(VariantID vid);
 
-  template < size_t block_size, size_t tune_idx >
+  template < size_t m_block_sz, size_t tune_idx >
   void runCudaVariantImpl(VariantID vid);
-  template < size_t block_size, size_t tune_idx >
+  template < size_t m_block_sz, size_t tune_idx >
   void runHipVariantImpl(VariantID vid);
   template < size_t work_group_size, size_t tune_idx >
   void runSyclVariantImpl(VariantID vid);
```
MrBurmark: You probably want to split this into a function for each of the tunings as the block_m tuning doesn't really need the `block_size` or `tune_idx` template params.

src/apps/LTIMES-Cuda.cpp
```diff
@@ -172,31 +246,114 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
...
+          RAJA::LaunchPolicy<RAJA::cuda_launch_t<async, factorized_m_block_sz*factorized_g_block_sz*factorized_z_block_sz>>;
 
-      using z_policy = RAJA::LoopPolicy<RAJA::cuda_global_size_z_loop<z_block_sz>>;
+      using z_policy = RAJA::LoopPolicy<RAJA::cuda_global_size_z_loop<factorized_z_block_sz>>;
 
-      using g_policy = RAJA::LoopPolicy<RAJA::cuda_global_size_y_loop<g_block_sz>>;
+      using g_policy = RAJA::LoopPolicy<RAJA::cuda_global_size_y_loop<factorized_g_block_sz>>;
 
-      using m_policy = RAJA::LoopPolicy<RAJA::cuda_global_size_x_loop<m_block_sz>>;
+      using m_policy = RAJA::LoopPolicy<RAJA::cuda_global_size_x_loop<factorized_m_block_sz>>;
```
MrBurmark: These should be direct policies.

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: Applied in f52fcfe0

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: Yes I applied a change to all the macros to match the naming convention change we made for the tunings 44686fa3

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: Fixed in 8518f96e

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: Wait... I reverted https://github.com/llnl/RAJAPerf/commit/f52fcfe0cb9aa6f6ed33d8d491df3d6fdf0c4c7b because we want the block size to be `m` for the kripke tuning. This change sets the compile time block size of 25 always.

src/apps/LTIMES.hpp
michaelmckinsey1: removed in b198051e

src/apps/LTIMES.hpp
michaelmckinsey1: Do you mean like 6045121e ?

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: a87ee04d

src/apps/LTIMES-Cuda.cpp
```diff
@@ -230,24 +380,68 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
 }
 
 
+void LTIMES::runCudaVariantM(VariantID vid)
```
michaelmckinsey1, line 384: Name `runCudaVariantM` combine with `runCudaVariantLaunchM`, so there is only one function templated on block size

michaelmckinsey1: d084a9f5

src/apps/LTIMES-Sycl.cpp
```diff
@@ -27,14 +27,13 @@ using namespace ltimes_idx;
 //
 // Define work-group shape for SYCL execution
 //
-#define m_wg_sz (32)
-#define g_wg_sz (integer::greater_of_squarest_factor_pair(work_group_size/m_wg_sz))
-#define z_wg_sz (integer::lesser_of_squarest_factor_pair(work_group_size/m_wg_sz))
+#define g_wg_sz (1)
```
MrBurmark: No zgm version for sycl?

michaelmckinsey1: [cc085af](https://github.com/llnl/RAJAPerf/pull/684/commits/cc085af89a077c1bb7a52dbe552b8c39b0ab9c38)

rhornung67:
@rchen20 when you have a minute please review this. Thank you.

src/apps/LTIMES-Sycl.cpp
```diff
@@ -27,14 +27,21 @@ using namespace ltimes_idx;
 //
 // Define work-group shape for SYCL execution
 //
-#define m_wg_sz (32)
-#define g_wg_sz (integer::greater_of_squarest_factor_pair(work_group_size/m_wg_sz))
-#define z_wg_sz (integer::lesser_of_squarest_factor_pair(work_group_size/m_wg_sz))
+#define m_g_wg_sz (1)
+#define m_z_wg_sz (1)
```
MrBurmark: Do we need to name these?

michaelmckinsey1: No we don't I have removed them

src/apps/LTIMES-Cuda.cpp
```diff
@@ -27,27 +27,50 @@ using namespace ltimes_idx;
...
+
+
+__global__ void ltimes_block_moments(PHI_VIEW phi, ELL_VIEW ell, PSI_VIEW psi,
+                       ID num_d, IM num_m, IG num_g, IZ num_z)
+{
+   IG g(blockIdx.y);
+   IZ z(blockIdx.x);
+
+   if (g < num_g && z < num_z) {
+     for (IM m(threadIdx.x); m < num_m; m += blockDim.x) {
```
MrBurmark: We probably want to explicitly hoist the `threadIdx` and `blockDim` constants into local variables.

src/apps/LTIMES-Cuda.cpp
MrBurmark: This is something that is worth looking into, but let's not increase the number of tunings further in this PR.

src/apps/LTIMES-Cuda.cpp
```diff
@@ -159,7 +223,6 @@ void LTIMES::runCudaVariantImpl(VariantID vid)
       startTimer();
       // Loop counter increment uses macro to quiet C++20 compiler warning
       for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
-
```
MrBurmark, line 162: leaving some whitespace in would be nice.

src/apps/LTIMES.hpp
```diff
@@ -114,17 +116,35 @@ class LTIMES : public KernelBase
...
+  void runHipVariantM(VariantID vid);
+  template < size_t tune_idx, size_t block_size >
+  void runHipVariantZGM(VariantID vid);
+
+  void runSyclVariantM(VariantID vid);
+  void runSyclVariantLaunchM(VariantID vid);
+  template < size_t work_group_size >
+  void runSyclVariantZGM(VariantID vid);
+  template < size_t work_group_size >
+  void runSyclVariantLaunchZGM(VariantID vid);
```
MrBurmark: Do you want to have named non-launch and launch helpers for all the gpu backends?

rhornung67 (commented):
@lc-hubcast approve

src/apps/LTIMES.hpp
michaelmckinsey1: [c40d485](https://github.com/llnl/RAJAPerf/pull/684/commits/c40d4859a57491b5b90cd393c55110fad1bf62b8)
Right we already decided to **not** do this for cuda/hip and I reintroduced it for sycl... fixed now

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: done in multiple places [cbd66af](https://github.com/llnl/RAJAPerf/pull/684/commits/cbd66af1b4f3908283bfb4333436de51a0d44b3a)

src/apps/LTIMES-Cuda.cpp
michaelmckinsey1: [c4d0702](https://github.com/llnl/RAJAPerf/pull/684/commits/c4d0702a893f29d7f419dd0d23f47702cdca3a96)

src/apps/LTIMES-Hip.cpp
```diff
@@ -27,26 +27,51 @@ using namespace ltimes_idx;
...
 
-#define LTIMES_NBLOCKS_HIP \
-  dim3 nblocks(static_cast<size_t>(RAJA_DIVIDE_CEILING_INT(*num_m, m_block_sz)), \
-               static_cast<size_t>(RAJA_DIVIDE_CEILING_INT(*num_g, g_block_sz)), \
-               static_cast<size_t>(RAJA_DIVIDE_CEILING_INT(*num_z, z_block_sz)));
+#define LTIMES_ZGM_THREADS_PER_BLOCK_TEMPLATE_PARAMS_HIP \
+  zgm_m_block_sz, zgm_g_block_sz, zgm_z_block_sz
 
+#define LTIMES_ZGM_THREADS_PER_BLOCK_HIP \
+  dim3 nthreads_per_block(LTIMES_ZGM_THREADS_PER_BLOCK_TEMPLATE_PARAMS_HIP);
```
rhornung67, line 46: There is a static assert after the similar line in the CUDA file here: https://github.com/llnl/RAJAPerf/pull/684/changes#diff-2ba93ddd312b66198ba5dd511529a6ba450079fb9d9b2dce9f223ce8ba5e95b8R47

Should that static assert be added here to be consistent?

michaelmckinsey1: [8f97248](https://github.com/llnl/RAJAPerf/pull/684/commits/8f97248d1823d0eb47f82f00f4497087fce4b43e) It was also missing in develop, so I added it here and in a couple of other missing places for consistency.

rhornung67 (approved):

rchen20 (approved):
Looks good overall! I'll note that LTimes and other Kripke kernels are not highly optimized, but these settings and the mirroring of Kripke LTimes performance is a good baseline. Thanks for doing this.

rhornung67 (commented):
@lc-hubcast approve

rhornung67:
@MrBurmark if you are good with this, please approve. Thank you.

MrBurmark (approved):

PR #693, opened 2026-06-19
`MASS3DPA` element batching

michaelmckinsey1:
# Summary

- This PR is an update for MASS3DPA
- It updates the `MASS3DPA` implementation to use element batching as was done in the reference implementation a few months ago https://github.com/mfem/mfem/pull/5299
- Testing `RAJA_CUDA` vs Laghos with mfem+raja, the runtimes for `MASS3DPA` and `SmemPAMassApply3D_Element` are within 1% runtime difference, however MASS3DPA is doing 10% less instructions.
- `RAJA_HIP` is 1% runtime difference between MASS3DPA and `SmemPAMassApply3D_Element` at problem size 134M (16M epm in Laghos)
- [x] Need to update/test the HIP version
- [x] Test Base_CUDA
- [x] Update MASSVEC3DPA, MASS3DPA-Atomic?
   - For MASS3DPA_ATOMIC and MASSVEC3DPA, the current kernels already use Q1D x Q1D x Q1D threads. Their threadIdx.z is real element-local work: dz or qz. So there is no spare z dimension to directly reinterpret as zbatch.
- [x] Batch size tunings are based on `RAJA_PERFSUITE_GPU_BLOCKSIZES`. So if `-DRAJA_PERFSUITE_GPU_BLOCKSIZES=16,32,64` then `block_64` (batch 16), `block_32` (batch 8), and `block_16` (batch 4)
   - ~linear: `block_64_d2q2t16`~
   - ~quadratic: `block_32_d4q4t2`~
   - ~cubic: `block_36_d6q6t1`~
- [x] Change tunings to match `RAJA_PERFSUITE_GPU_BLOCKSIZES`


src/apps/MASS3DPA.hpp
```diff
@@ -172,8 +172,22 @@
...
+//   constexpr RAJA::Index_type Q1D = 4;
+// cubic
+// constexpr RAJA::Index_type D1D = 6;
+//   constexpr RAJA::Index_type Q1D = 6;
+  constexpr RAJA::Index_type D1D = 2;
+  constexpr RAJA::Index_type Q1D = 2;
+
+  constexpr RAJA::Index_type TBATCH = 16; // linear
+  // constexpr RAJA::Index_type TBATCH = 2; // quadratic
+  // constexpr RAJA::Index_type TBATCH = 1; // cubic
```
michaelmckinsey1: This is my observation of the values via Laghos. Looking for feedback here and whether we should try to have different configurations of MASS3DPA for different orders?

MrBurmark: First what do we want to cover? If we want to do more, I have thought about templating entire kernels to handle things like different orders and maybe using explicit template instantiation to keep the instantiation in separate files similar.

artv3:
Nice @michaelmckinsey1 ! What would be neat is to add a batching parameter and allow for different batch sizes, maybe 4 is good GPU X or 7 is good for GPU Y type of thing

MrBurmark:
> Nice @michaelmckinsey1 ! What would be neat is to add a batching parameter and allow for different batch sizes, maybe 4 is good GPU X or 7 is good for GPU Y type of thing

I think he means different batch sizes as tunings.

artv3:
> > Nice @michaelmckinsey1 ! What would be neat is to add a batching parameter and allow for different batch sizes, maybe 4 is good GPU X or 7 is good for GPU Y type of thing
> 
> I think he means different batch sizes as tunings.

yes, sounds good to me!

src/apps/MASS3DPA-Cuda.cpp
```diff
@@ -26,64 +26,91 @@ namespace apps {
...
-  MASS3DPA_0_GPU
+  MASS3DPA_GPU_SMEM_DECL(mpa::TBATCH)
+  MASS3DPA_GPU_SMEM_SLICE(zbatch)
 
   GPU_FOREACH_THREAD_INC(dy, y, mpa::D1D, mpa::Q1D) {
     GPU_FOREACH_THREAD_INC(dx, x, mpa::D1D, mpa::Q1D){
-      MASS3DPA_1
+      if (valid_e) {
+        MASS3DPA_1
+      }
```
helloworld922: Instead of doing the check for `valid_e` inside of every loop what happens if you do a single check `if(!valid_e) return;` at the beginning? That's how we have it implemented in MFEM.

src/apps/MASS3DPA.hpp
helloworld922: It would also be helpful to see how the performance of quadratic compares with and without RAJA since that's what MARBL mainly uses.

src/apps/MASS3DPA-Cuda.cpp
```diff
@@ -26,64 +26,91 @@ namespace apps {
...
   }
+  if (threadIdx.z == 0) {
+    GPU_FOREACH_THREAD_INC(dy, y, mpa::D1D, mpa::Q1D) {
+      GPU_FOREACH_THREAD_INC(dx, x, mpa::Q1D, mpa::Q1D) {
+        MASS3DPA_2
+      }
+    }
+  }
   __syncthreads();
   GPU_FOREACH_THREAD_INC(dy, y, mpa::D1D, mpa::Q1D) {
```
helloworld922: It would be interesting to compare having this be a for loop vs. doing a simple if statement check (`blockDim.x` and `blockDim.y` should always be >= `Q1D`).

This seems to help with the hip compiler.

src/apps/MASS3DPA-Cuda.cpp
michaelmckinsey1: See [efaa790](https://github.com/llnl/RAJAPerf/pull/693/commits/efaa7903c014434c48690b4b71453d87f3fccfa1)

src/apps/MASS3DPA-Cuda.cpp
michaelmckinsey1: Are you proposing to only implement this check as an `if` in the RAJAPerf kernel and not in the mfem kernel? I think we want to keep the implementations similar. If the mfem kernel is general then this should be general.

Unless I am mistaken, but it appears the mfem kernel is also a for loop here https://github.com/mfem/mfem/blob/17d1afc3b78c55b17316dc262604a83bd91364d0/fem/integ/bilininteg_mass_kernels.hpp#L872

helloworld922: I think it would be interesting to have a way to test and compare the performance of both. We've noticed that in some situations particularly on hip that the for loop causes the compiler to introduce a bunch of synchronization statements which significantly reduces performance.

src/apps/MASS3DPA-Cuda.cpp
```diff
@@ -267,7 +393,28 @@ void MASS3DPA::runCudaVariantImpl(VariantID vid) {
   }
 }
 
-RAJAPERF_GPU_BLOCK_SIZE_TUNING_DEFINE_BOILERPLATE(MASS3DPA, Cuda, Base_CUDA, RAJA_CUDA)
```
michaelmckinsey1, line 270: I can't use the boilerplate because runCudaVariantImpl is templated from d1d, q1d, and TBATCH

src/apps/MASS3DPA.cpp
```diff
@@ -46,49 +78,62 @@ MASS3DPA::MASS3DPA(const RunParams& params)
...
+  const Size_type y_len = m_D1D*m_D1D*m_D1D*m_NE;
+  const Checksum_type y_checksum =
+      calcChecksum(getDataSpace(vid), m_Y, y_len, getDataAlignment());
+  const Checksum_type x_checksum =
+      calcChecksum(getDataSpace(vid), m_X, y_len, getDataAlignment());
+  const Checksum_type expected_value =
+      Checksum_type(m_D1D) * Checksum_type(m_D1D) * Checksum_type(m_D1D) *
+      Checksum_type(m_Q1D) * Checksum_type(m_Q1D) * Checksum_type(m_Q1D);
+
+  addToChecksum(y_checksum / (x_checksum * expected_value));
```
michaelmckinsey1: @artv3 I am normalizing the checksum so the quadratic and cubic tunings do not fail the checksum test. Not sure how else they would be comparable.

artv3: what is the reference solution? I think we would want to compare same orders

src/apps/MASS3DPA.hpp
michaelmckinsey1: @helloworld922 The largest quadratic size I have ran at the moment is NE=2M, and I see RAJA only being 1% slower compared to Base (1.98ms vs 2ms)

src/apps/MASS3DPA.cpp
MrBurmark: Things that change the problem itself and hence things like the number of flops, or the amount of memory used have to be separate kernels. So quadratic and cubic should be separate kernels.

michaelmckinsey1:
I need to revert the order tuning changes. This would need to be implemented as separate files. As is, the FLOP and bytes counts are incorrect. I will make different tunings based on batch size only

src/apps/MASS3DPA-Cuda.cpp
```diff
@@ -267,7 +393,24 @@ void MASS3DPA::runCudaVariantImpl(VariantID vid) {
...
```
michaelmckinsey1, line 270: @MrBurmark 
This could be reverted back to using the boilerplate if I change the tuning name back to `block_X` instead of `block_X_batch_Y`.

Should I revert?

MrBurmark: Sure, if you don't mind that the batch size information is lost.

artv3:
@artv3 : please review

src/apps/MASS3DPA-Hip.cpp
```diff
@@ -124,129 +166,221 @@ void MASS3DPA::runHipVariantImpl(VariantID vid) {
...
+                            MASS3DPA_2
+                          }
+                        );  // RAJA::loop<inner_x>
+                      } // lambda (dy)
+                    );  // RAJA::loop<inner_y>
+                  }
+                }
+              );  // RAJA::loop<inner_z>
+
+              RAJA::loop<inner_z>(ctx, RAJA::RangeSegment(0, 1),
```
artv3: RAJA recently release the ``RAJA::mask`` to do this kind of operation: 
https://github.com/llnl/RAJA/blob/develop/examples/raja-launch.cpp#L167 
Could you use that instead? We can update in other places of the suite in a different PR.

artv3 (commented):
Since many loops start at 0, it might be nice to also use ``RAJA::range(N)`` for compactness, its new so this would be a first kernel to adopt it if you choose.

michaelmckinsey1:
@artv3 I have applied those changes, although [8b2cabc](https://github.com/llnl/RAJAPerf/pull/693/commits/8b2cabcb54f83d8939fae410270500d79306ed03) requires making changes to the `-Seq` and `-OMP` variants to be uniform across the kernel (I wasn't touching these previously in this PR). 

Also, a performance concern for [8b2cabc](https://github.com/llnl/RAJAPerf/pull/693/commits/8b2cabcb54f83d8939fae410270500d79306ed03). I ran a really simple A/B with and without this commit and the `Mean time per rep (sec.)` across 50 reps was `0.00231193` vs `0.00365900` which is 1.58x slower. I don't see why this change would be so significant, but this is concerning. I assume this is unexpected? This was on one H100, one rank, `178956968` problem size.

[7d1493b](https://github.com/llnl/RAJAPerf/pull/693/commits/7d1493b69aff00e116b6c3bc158c4da220d54d29) has 0 impact on performance (<0.1%).

artv3:
> 8b2cabc

That is very unexpected given that the return type is just a RAJA::TypedRangeSegement , can you make an issue in RAJA? I can dig in to that when I get a chance

michaelmckinsey1:
> > [8b2cabc](https://github.com/llnl/RAJAPerf/commit/8b2cabcb54f83d8939fae410270500d79306ed03)
> 
> That is very unexpected given that the return type is just a RAJA::TypedRangeSegement , can you make an issue in RAJA? I can dig in to that when I get a chance

I've created https://github.com/llnl/RAJA/issues/2087

MrBurmark:
> > > [8b2cabc](https://github.com/llnl/RAJAPerf/commit/8b2cabcb54f83d8939fae410270500d79306ed03)
> > 
> > 
> > That is very unexpected given that the return type is just a RAJA::TypedRangeSegement , can you make an issue in RAJA? I can dig in to that when I get a chance
> 
> I've created [llnl/RAJA#2087](https://github.com/llnl/RAJA/issues/2087)

This performance regression should be fixed by https://github.com/llnl/RAJA/pull/2089.

rhornung67:
@artv3 @MrBurmark is it OK to merge this, or should we wait until we pull in https://github.com/llnl/RAJA/pull/2089

MrBurmark:
I suppose its not a big deal as long as the RAJA update follows fairly soon.

rhornung67:
> I suppose its not a big deal as long as the RAJA update follows fairly soon.

The RAJA update PR is set to automerge. When that goes in, I will update this PR and set on automerge. We need someone to approve. @artv3 or @MrBurmark ?

MrBurmark (approved):

michaelmckinsey1:
> > > > [8b2cabc](https://github.com/llnl/RAJAPerf/commit/8b2cabcb54f83d8939fae410270500d79306ed03)
> > > 
> > > 
> > > That is very unexpected given that the return type is just a RAJA::TypedRangeSegement , can you make an issue in RAJA? I can dig in to that when I get a chance
> > 
> > 
> > I've created [llnl/RAJA#2087](https://github.com/llnl/RAJA/issues/2087)
> 
> This performance regression should be fixed by [llnl/RAJA#2089](https://github.com/llnl/RAJA/pull/2089).

2089 did fix the performance issue (https://github.com/llnl/RAJA/issues/2087#issuecomment-5764605602)

src/apps/MASS3DPA.hpp
artv3: Throughput studies are the best, for applications a single point can be tough as the size of the problem tends to be limited by what is enabled in the code. Check out the compiler matrix script for generating throughput studies: https://github.com/llnl/RAJAPerf/blob/develop/scripts/build_compiler_matrix.py

rhornung67 (commented):
@lc-hubcast approve

rhornung67 (approved):

PR #695, opened 2026-06-24
Caliper: Annotate kernels separately

michaelmckinsey1:
- [x] Depends on #700 

# Summary

- This PR is additional Caliper regions for RAJAPerf kernels that have multiple kernel launches.
  - Adds additional Caliper regions at the request of @pearce8 
- There is currently no synchronize for these regions for asynchronous GPU kernels. But these should be profiled with CUDA/HIP events anyway, which do not need CPU synchronization to measure GPU time.
- [x] `RAJA::launch` or `RAJA::forall` kernels are annotated using the RAJA Caliper plugin
- [x] add function types to be able to filter out kernels instead of `type=function` -> `type=subkernel` [1]
- [x] Update documentation for naming convention

## Examples

`Polybench_JACOBI_1D` has one launch per rep for `poly_jacobi_1D_1` and one for `poly_jacobi_1D_2`. So its tree will now profile each separately:
image

`POLYBENCH_FLOYD_WARSHALL`, `HALO_PACKING`, `HALO_EXCHANGE` all have variable amount of launches, so we will append `_k` instead of adding k regions to the tree
image

image

image

All kernels with only 1 launch per rep are unchanged.

## [1] how to filter `subkernel`s at runtime

`raja-perf.exe --atcc "exclude_attributes=subkernel"`

Then subkernels will be excluded:
image

_Caveat for this method is that if profiling API functions, like HIP or MPI the subkernel regions will no longer be excluded if they have API functions._


michaelmckinsey1:
made a draft because need to add different types to these regions so we can filter them with caliper

src/apps/ENERGY-Cuda.cpp
```diff
@@ -202,35 +214,47 @@ void ENERGY::runCudaVariantImpl(VariantID vid)
...
 
+        RP_CALI_MARK_BEGIN(RP_CALI_REGION(ENERGY_1));
         RAJA::forall< RAJA::cuda_exec<block_size, async> >( res,
           RAJA::RangeSegment(ibegin, iend), [=] __device__ (Index_type i) {
           ENERGY_BODY1;
         });
+        RP_CALI_MARK_END(RP_CALI_REGION(ENERGY_1));
 
+        RP_CALI_MARK_BEGIN(RP_CALI_REGION(ENERGY_2));
         RAJA::forall< RAJA::cuda_exec<block_size, async> >( res,
```
artv3, line 225: @michaelmckinsey1 , why not use just use RAJA's kernel naming capability here? Example... 
```
RAJA::forall<RAJA::cuda_exec<256>>(
    range, 
    RAJA::Name("VectorAddKernel"), // <-- Kernel Name injected here
    [=] RAJA_DEVICE (int i) {
        c[i] = a[i] + b[i];
    }
);
````

Comment (michaelmckinsey1) [reply src/apps/ENERGY-Cuda.cpp:225]: Wouldn't this only work for RAJA variants? Or is that what you are proposing, i.e., annotate base and RAJA variants differently.

Comment (artv3) [reply src/apps/ENERGY-Cuda.cpp:225]: yes only for RAJA variants

Comment (michaelmckinsey1) [reply src/apps/ENERGY-Cuda.cpp:225]: Done now

Comment (michaelmckinsey1) [reply src/apps/ENERGY-Cuda.cpp:225]: @artv3 I'm actually going to revert this, and not use the plugin for two reasons:
1. To keep the RAJA and non-RAJA (e.g. Base) annotations consistent. As it is right now the RAJA kernels use the plugin, where everything else is using the macro annotations.
2. To enable the Caliper subkernel filtering as shown in https://github.com/llnl/RAJAPerf/pull/695/changes#diff-f1b7ede2b59bf264efbe43bc92349deb28c8f49dc87efa11b4882ab6805beb24R665, which is defined in my macros.

Comment (MrBurmark) [reply src/apps/ENERGY-Cuda.cpp:225]: Since we want to be able to compile these out and perhaps add additional features later I think this makes sense.

**Review by MrBurmark — APPROVED** [review]

**Review by artv3 — APPROVED** [review]

**Review by rhornung67 — APPROVED** [review]
Thank you @michaelmckinsey1

**Comment by michaelmckinsey1** [comment]
@MrBurmark @artv3 @rhornung67 
I added one new commit since this was approved [adebe6c](https://github.com/llnl/RAJAPerf/pull/695/commits/adebe6c850f9ee43f56afe3b3c178acde5969c73). I noticed while testing for many different variants there were missing annotations. For example, the RAJA variant would be annotated but not Base and Lambda. This commit did not annotate any new kernels. This happened because I was using an LLM to apply the annotations and apparently, despite giving manual examples and describing that there are different variants, it has a really difficult time identifying that the annotations should be applied uniformly across the variants (this surprised me since I thought it would be trivial). 

Also it is now rebased on #700

File: docs/sphinx/dev_guide/kernel_class_impl.rst [hunk 2]
```diff
@@ -286,6 +286,23 @@ A few details are worth noting:
     compilers from optimizing certain loop execution operations. The syntax
     inside the macro definition quiets compiler warnings when the code is 
     compiled using the C++20 standard.
+  * Kernels that execute multiple meaningful loop bodies or launch multiple
```
MrBurmark: Reminder to update docs, if we're marking all subkernels now.

docs/sphinx/dev_guide/kernel_class_impl.rst
```diff
@@ -286,6 +286,23 @@ A few details are worth noting:
...
+    kernels per repetition should mark each subkernel with
+    ``RP_CALI_SUBKERNEL_BEGIN(RP_CALI_REGION(<name>))`` and
```
MrBurmark: RP_CALI_REGION just stringifies its input, why not just use string literals?

michaelmckinsey1: I think there was some reason for this initially, but not anymore. I'm using string literals now.

docs/sphinx/dev_guide/kernel_class_impl.rst
michaelmckinsey1: done

MrBurmark (approved):

PR #696, opened 2026-07-15
Femsweep cuda fix

stingyemperor:
# Summary

- This PR is a fix for the CUDA version of femsweep to bring its performance closer to the HIP version
- It does the following:
  - Adds loop unrolling to femsweep as requested by @pearce8 

# Details 
We found that the performance of the femsweep kernel was around 2-3x worse on NVIDIA(H100) vs AMD(MI300A). The NVIDIA version was spilling to local memory instead of using registers and forcing loops to unroll fixed the issue.

Note: This version also changes the block size for the NVIDIA version to be 64 to match the AMD version.

### DIFF
--- a/src/apps/FEMSWEEP-Cuda.cpp
+++ b/src/apps/FEMSWEEP-Cuda.cpp
@@ -7,7 +7,6 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
 
-#define FEMSWEEP_ENABLE_UNROLL
 #include "FEMSWEEP.hpp"
 
 #include "RAJA/RAJA.hpp"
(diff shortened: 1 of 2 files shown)


src/apps/FEMSWEEP.hpp
```diff
@@ -96,6 +96,20 @@ constexpr long ND = 8;   // number of corners per element
 constexpr long NLF = 6;  // number of faces per element
 constexpr long FDS = 4;  // number of DOFs per face
 
+#if defined(FEMSWEEP_ENABLE_UNROLL)
+#define FEMSWEEP_UNROLL _Pragma("unroll")
```
MrBurmark: Won't this fully unroll things anyway, why have different macros for different sizes?

stingyemperor: Yes it will, I had it separate to test if there was a better unroll count. But, I can take it out to make things cleaner.

MrBurmark:
Does it make more sense to have a tuning for unrolled so we can compare what the compiler does by default and what it does when told to unroll everything?

src/apps/FEMSWEEP.hpp
artv3: RAJA has some unroll macros: 
https://github.com/llnl/RAJAPerf/blob/338cfeb7266607648bd06815ba103f61e357c2e5/src/apps/FEM_MACROS.hpp#L14-L19 

[include/RAJA/config.hpp.in](https://github.com/llnl/RAJA/blob/12bc00ee8a0649876d58d1daa4b11c6c38cdf165/include/RAJA/config.hpp.in#L447)

Maybe these could be applied?

rhornung67: I think that's preferable since it gives the ability to try different unroll factors.

MrBurmark: I think we probably just want to fully unroll here, that's what allows the compiler to see that the array indexing is static instead of dynamic.

MrBurmark:
I think this is conflicting with #695

MrBurmark:
@rchen20 do we want a cuda and a hip version of the unrolled tuning

MrBurmark:
@stingyemperor do we want to add other backend unroll versions like hip?

rhornung67:
@rchen20 please take a look at this one too. Thank you.

src/apps/FEMSWEEP-Cuda.cpp
```diff
@@ -22,153 +22,188 @@ namespace rajaperf
...
-      }
-      stopTimer();
+FEMSWEEP_DEFINE_CUDA_KERNEL(FEMSweep3D, FEMSWEEP_KERNEL_HYPERPLANE_ELEMENT)
+FEMSWEEP_DEFINE_CUDA_KERNEL(FEMSweep3DUnroll, FEMSWEEP_KERNEL_HYPERPLANE_ELEMENT_UNROLL)
+
+#undef FEMSWEEP_DEFINE_CUDA_KERNEL
+
+#define FEMSWEEP_DEFINE_CUDA_VARIANT_IMPL(METHOD_NAME, KERNEL_NAME, ELEMENT) \
+template < size_t block_size > \
+void FEMSWEEP::METHOD_NAME(VariantID vid) \
```
MrBurmark, line 70: This uses more macros than even other areas of the code. Other areas use a template parameter and an if constexpr to decide between different similar implementations.

rchen20: Agreed, I will un-macro all of this.

src/apps/FEMSWEEP.hpp
```diff
@@ -296,13 +317,15 @@ class FEMSWEEP : public KernelBase
...
   template < size_t block_size >
+  void runCudaVariantImplUnroll(VariantID vid);
+  template < size_t block_size >
   void runHipVariantImpl(VariantID vid);
 
 private:
 #if defined(RAJA_ENABLE_HIP)
   static const size_t default_gpu_block_size = 64;
 #else
-  static const size_t default_gpu_block_size = 128;
```
rchen20, line 305: Curious, why was this decreased? Were there occupancy issues?

MrBurmark: This was done to make the cuda and hip backends match.

src/apps/FEMSWEEP-Cuda.cpp
```diff
@@ -75,14 +79,12 @@ void FEMSWEEP::runCudaVariantImpl(VariantID vid)
     case Base_CUDA : {
 
       startTimer();
-      // Loop counter increment uses macro to quiet C++20 compiler warning
       for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
 
-         RP_CALI_SUBKERNEL_BEGIN("FEMSWEEP_1");
```
MrBurmark, line 81: These should be here.

src/apps/FEMSWEEP.hpp
```diff
@@ -198,72 +202,154 @@ RAJA_HOST_DEVICE inline void SolveLinearSystemNxN(Real_ptr A,
...
+                               const Index_type g,
+                               const Index_type k,
+                               const Index_type nehp_pos,
+                               const Real_type Ffactor)
+{
+  Real_type A[ND*ND];
+  Real_type b[ND];
+  const Index_type e = order_r[k + nehp_pos + a * ne];
+
+  femsweepFor<unroll>(0, ND, [&](Index_type j)
```
MrBurmark, line 313: I think it would be better to have a macro where we either pass RAJA_UNROLL or an empty macro so there aren't lambdas in here.
```cpp

#define ROLL
#define UNROLL _Pragma("unroll")

#define KERNEL(ROLLING) \
  ROLLING \
  for (...) { ... }
  
#define KERNEL_ROLLED KERNEL(ROLL)
#define KERNEL_UNROLLED KERNEL(UNROLL)
```

src/apps/FEMSWEEP-Hip.cpp
```diff
@@ -52,7 +52,11 @@ __global__ void FEMSweep3D( const Real_ptr Bdat,
     const Index_type nehp = phpaa_r[ohp + hp];
     for (Index_type k = threadIdx.x; k < nehp; k += block_size)
     {
-      FEMSWEEP_KERNEL_HYPERPLANE_ELEMENT;
+      femsweepHyperplaneElement<false>(
```
MrBurmark, line 55: Did we want to have a hip version of the unrolled one for symmetry?

PR #699, opened 2026-08-04
CAMP sycl API changed from using sycl::queue ptr to reference; change…

smithsg84:
Changes sycl queue pointer to sycl queue reference.  CAMP changed so this follows CAMP and more closely matches the Sycl API (e.g. sycl allocation and free calls take reference).


adrienbernede:
I'm merging this since it passed the sycl test.

PR #702, opened 2026-08-07
Implementation of Basic_MAT_MAT, a non shared memory variant of Basic_MAT_MAT_SHARED

Yejashi:
# Summary

- Adds `Basic_MAT_MAT`, a non-shared-memory variant of `Basic_MAT_MAT_SHARED`.

# Motivation

In a performance study comparing NVIDIA and AMD GPUs on the same kernels, we observed a large NVIDIA/AMD performance gap for `Polybench_GEMM`, while the gap was much narrower for `Basic_MAT_MAT_SHARED`.

This PR adds `Basic_MAT_MAT` to isolate the effect of shared memory and compare the same tiled matrix-matrix computation without shared-memory staging.

PR #708, opened 2026-09-03
Impl of Basic_MAT_MAT which is Basic_MAT_MAT_SHARED without shared memory

rhornung67:
# Summary

- This PR is a copy of https://github.com/llnl/RAJAPerf/pull/702 which has the PR branch on a fork. 
- This companion PR was made so we can make updates to the PR branch.


MrBurmark (approved):

michaelmckinsey1 (approved):
As Befikir said, we have found this kernel useful for studying tiling without shared memory.

src/basic/MAT_MAT-Cuda.cpp
```diff
@@ -0,0 +1,209 @@
...
+    // Loop counter increment uses macro to quiet C++20 compiler warning
+    for (RepIndex_type irep = 0; irep < run_reps; RP_REPCOUNTINC(irep)) {
+
+      RP_CALI_SUBKERNEL_BEGIN("MAT_MAT_1");
+      RAJA::launch<launch_policy>( res,
+        RAJA::LaunchParams(RAJA::Teams(Nx, Ny),
+                         RAJA::Threads(tile_size, tile_size)),
+        [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+
+          RAJA::loop<teams_y>(ctx, RAJA::RangeSegment(0, Ny),
```
artv3, line 164: This is minor, but the new ``RAJA::range(Ny)`` allows users to skip the start index when it is zero. Allows for compactness and easy reading :)

artv3 (approved):

src/basic/MAT_MAT.hpp
```diff
@@ -0,0 +1,136 @@
...
+  Real_ptr C = m_C;
+
+//
+// Unlike MAT_MAT_SHARED, no RAJA_TEAM_SHARED arrays are declared, so there is
+// no BODY_0 and no need for the CLANG/HIP host-compile workaround.
+//
+#define MAT_MAT_BODY_1(tile_size)                                              \
+  const Index_type Row = by * tile_size + ty;                                  \
+  const Index_type Col = bx * tile_size + tx;                                  \
+  Real_type dot = 0.0;
```
artv3, line 75: One thing worth noting is that the shared memory version, does not have a thread specific accumulator due to portability challenges between the CPU and GPU.

PR #709, opened 2026-09-08
Fixing Kokkos kernels that produce incorrect checksums or access memory improperly

Yejashi:
# Summary

This PR fixes several Kokkos variants that were either producing incorrect checksums or accessing memory outside their allocations.

Changes include:

* `HYDRO_1D-Kokkos.cpp`

  * Fixes the Kokkos Views being created with `iend + 12` elements over `iend`-element allocations.
  * `m_z` is now wrapped from its actual allocation base (`z + 10`, extent `iend + 1`), and the kernel uses `z_view[i]` / `z_view[i + 1]`.

* `DIFF_PREDICT-Kokkos.cpp`

  * Fixes Views being created from pointers shifted before the beginning of `m_px` and `m_cx`.
  * The Views now wrap the actual allocations (`iend * 10` and `iend`), with the column indices adjusted accordingly.

* `NODAL_ACCUMULATION_3D-Kokkos.cpp`

  * Fixes five node pointers that did not match `NDPTRSET` in `AppsData.hpp`.
  * Also fixes `vol` using `m_nodal_array_length` instead of `m_zonal_array_length`.

* `ARRAY_OF_PTRS-Kokkos.cpp`

  * Removes an extra `a *` from the kernel body. `ARRAY_OF_PTRS_BODY` should just compute the sum.

* `PI_ATOMIC-Kokkos.cpp`

  * Writes the final result to `m_pi_final`, which is what `updateChecksum` uses.
  * Adds the missing `PI_ATOMIC_DATA_TEARDOWN` corresponding to the allocation in `PI_ATOMIC_DATA_SETUP`.

* `REDUCE3_INT-Kokkos.cpp`

  * Assigns the reduction result for each repetition instead of accumulating across repetitions, matching the Seq implementation.
  * Previously the checksum was effectively multiplied by `run_reps`.

* `POLYBENCH_HEAT_3D-Kokkos.cpp`

  * Moves the two `moveDataToHostFromKokkosView` calls outside the `switch`.
  * In the `Kokkos_Lambda` case, the `break` skipped those copies and the checksum was calculated from stale host data.

## Notes
I ran into issues with the aforementioned when buildingg and running RAJAPerf with Kokkos enabled on Tuolumne. 

## Testing
Tested with Serial, OpenMP, and HIP on gfx942 using ROCm 7.2.0 on Tuolumne.


src/lcals-kokkos/DIFF_PREDICT-Kokkos.cpp
```diff
@@ -52,25 +54,25 @@ void DIFF_PREDICT::runKokkosVariant(VariantID vid) {
...
-            ar = cr - px_view(i, 9);
-            px_view(i, 9) = cr;
-            br = ar - px_view(i, 10);
-            px_view(i, 10) = ar;
-            cr = br - px_view(i, 11);
-            px_view(i, 11) = br;
-            px_view(i, 13) = cr - px_view(i, 12);
-            px_view(i, 12) = cr;
+            ar = cx_view(i);
+            br = ar - px_view(i, 0);
```
MrBurmark, line 55: Is this offsetting now different than the offsetting done in the other variants?

Yejashi: Yes, the indexing representation is different in the Kokkos variant, but it maps to the same underlying elements as the other variants. The other variants shift px/cx back by 4 * offset and then compensate for that in the body with indices starting at  4.

MrBurmark:
That makes sense, I'm mainly worried that we're running a slightly different kernel in the kokkos version due to the differences in offset calculations. We could go back to the version of the kernel before I changed it that allocated more memory than it actually used. I'm a bit reluctant to do that as then the allocated memory metadata will be misleading as not all of that memory will be used.

src/lcals-kokkos/DIFF_PREDICT-Kokkos.cpp
MrBurmark: Ok, let's add a comment in the kokkos implementation that its doing something slightly different in how it calculates offsets. If we want to change that we can do that separately.

MrBurmark (approved):
@lc-hubcast approve

MrBurmark:
@Yejashi can you update this branch?

Yejashi:
> @Yejashi can you update this branch?

Updated, @MrBurmark.

rhornung67 (commented):
@lc-hubcast approved

PR #718, opened 2026-09-21
CUDA 13 Updates

rchen20:
# Summary

- This PR is a refactoring
- It does the following:
  - Updates cudamemadvise and CUB sum calls to the proper CUDA 13 API.


src/algorithm/REDUCE_SUM-Cuda.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "cub/device/device_reduce.cuh"
 #include "cub/util_allocator.cuh"
+#include <cuda/std/functional>
```
MrBurmark, line 20: I assume this works for cuda 11+?

rhornung67: We can't use CUDA 11 now that we are on C++20.

MrBurmark: Right it was cuda 12+, thanks.

rchen20: FYI, if we had to roll back to C++17 for whatever reason, this use of `cuda/std/functional` would still work with CUDA 11. `cuda::std::plus<>` has been available since CUDA 11, and our old `cub::Sum` was an alias to `cuda::std::plus<>` since CUDA 12.

rhornung67 (commented):
@lc-hubcast approve

src/common/CudaDataUtils.hpp
```diff
@@ -145,6 +145,41 @@ inline int getCudaDevice()
...
+ * \brief Apply CUDA memory advice for the current device.
+ */
+inline void adviseCudaMemoryForDevice(const void* ptr, Size_type len, cudaMemoryAdvise advice)
+{
+  const int device = getCudaDevice();
+#if CUDART_VERSION >= 13000
+  cudaMemLocation devLoc {};
+  devLoc.type = cudaMemLocationTypeDevice;
+  devLoc.id = device;
+  CAMP_CUDA_API_INVOKE_AND_CHECK( cudaMemAdvise,
```
adayton1, line 158: Do we lose some info now (like file and line number of actual call site)? If so, it may be worth considering using std::source_location: https://en.cppreference.com/cpp/utility/source_location.

adayton1 (approved):

MrBurmark (approved):

src/algorithm/REDUCE_SUM-Cuda.cpp
MrBurmark: Thanks for the info.

src/common/CudaDataUtils.hpp
MrBurmark: I think we can defer that to a separate MR.

rhornung67 (approved):

